PR #1, opened 2021-01-29
fix buildroot crosscompile

pseiderer:
Fix some findings while preparing a buildroot libcamera-apps package:

- cmake: remove unsafe host include path for libdrm (use pkgconfig instead)
- cmake: remove unrecognized gcc command-line option
- libcamera_vid: fix getline call (needs size_t)
- libcamera_still: fix getline call (needs size_t)
- libcamera_app: fix int64_t constants
- cmake: fix linking with exif, jpeg
- cmake: add install target
- cmake: no need to link against ipa_rpi.so

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,7 +13,13 @@ set (CMAKE CXX_FLAGS "-Wall -Wextra -pedantic -Wno-unused-parameter -faligned-ne
 add_definitions(-Wfatal-errors)
 add_definitions(-Wno-psabi)
 add_definitions(-DBOOST_LOG_DYN_LINK)
-add_definitions(-mfpu=neon-fp-armv8 -ftree-vectorize)
+
+IF (NOT ENABLE_COMPILE_FLAGS_FOR_TARGET)
+  set(ENABLE_COMPILE_FLAGS_FOR_TARGET "armv8")
+endif()
+if ("${ENABLE_COMPILE_FLAGS_FOR_TARGET}" STREQUAL "armv8")
+  add_definitions(-mfpu=neon-fp-armv8 -ftree-vectorize)
+endif()
 
 project(common)
 add_library(common output.cpp file_output.cpp net_output.cpp circular_output.cpp egl_preview.cpp drm_preview.cpp jpeg.cpp)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,31 +50,28 @@ message(STATUS "${EPOXY_LIBRARY}")
 find_library(DRM_LIBRARY libdrm.so REQUIRED)
 message(STATUS "${DRM_LIBRARY}")
 
-find_library(IPA_RPI_LIBRARY ipa_rpi.so PATHS "${CAMERA_LIBRARY_DIRS}/libcamera" REQUIRED)
-message(STATUS "${IPA_RPI_LIBRARY}")
-
 add_executable(libcamera-still libcamera_still.cpp yuv.cpp dng.cpp png.cpp bmp.cpp)
-target_link_libraries(libcamera-still common "${LIBCAMERA_LIBRARY}" "${IPA_RPI_LIBRARY}" exif ${Boost_LIBRARIES} jpeg tiff png pthread "${X11_LIBRARIES}" "${EPOXY_LIBRARY}" "${DRM_LIBRARY}")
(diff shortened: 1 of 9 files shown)


CMakeLists.txt
```diff
@@ -13,7 +13,6 @@ set (CMAKE CXX_FLAGS "-Wall -Wextra -pedantic -Wno-unused-parameter -faligned-ne
 add_definitions(-Wfatal-errors)
 add_definitions(-Wno-psabi)
 add_definitions(-DBOOST_LOG_DYN_LINK)
-add_definitions(-mfpu=neon-fp-armv8 -ftree-vectorize)
```
davidplowman, line 16: I'd quite like to like these compile flags in place for platforms that support them. Do you know of a way to do that? Thanks!

CMakeLists.txt
```diff
@@ -43,28 +43,28 @@ message(STATUS "${EPOXY_LIBRARY}")
...
+target_link_libraries(libcamera-raw encoders common "${LIBCAMERA_LIBRARY}" exif jpeg ${Boost_LIBRARIES} pthread "${X11_LIBRARIES}" "${EPOXY_LIBRARY}" "${DRM_LIBRARY}")
 
 project(libcamera-jpeg)
 
 add_executable(libcamera-jpeg libcamera_jpeg.cpp)
-target_link_libraries(libcamera-jpeg common "${LIBCAMERA_LIBRARY}" "${IPA_RPI_LIBRARY}" exif jpeg ${Boost_LIBRARIES} pthread "${X11_LIBRARIES}" "${EPOXY_LIBRARY}" "${DRM_LIBRARY}")
+target_link_libraries(libcamera-jpeg common "${LIBCAMERA_LIBRARY}" exif jpeg ${Boost_LIBRARIES} pthread "${X11_LIBRARIES}" "${EPOXY_LIBRARY}" "${DRM_LIBRARY}")
+
+install(TARGETS common encoders LIBRARY)
+install(TARGETS libcamera-still libcamera-vid libcamera-hello libcamera-raw libcamera-jpeg RUNTIME)
```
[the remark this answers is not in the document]
pseiderer: According to [1] the defaults depend on the cmake version, but the following should fix it (see [2]):

```
install(TARGETS common encoders LIBRARY DESTINATION lib ARCHIVE DESTINATION lib)
install(TARGETS libcamera-still libcamera-vid libcamera-hello libcamera-raw libcamera-jpeg RUNTIME DESTINATION bin)
```

[1] https://stackoverflow.com/questions/14990343/cmake-error-targets-given-no-library-destination-for-shared-library-target
[2] https://cmake.org/cmake/help/latest/command/install.html

CMakeLists.txt
pseiderer: Sorry, no easy way (and the buildroot view/preferred way is to avoid package specific  heuristics to determine architecture specific compile flags - in buildroot the preferred way is build all packages specific to your selected target architecture/cpu etc. by a global configure/selection).

One problem is (for cross compiling) the detection of the target arch and/or the suitable/supported compiler flags (and not all cross-compiler available flags/features will be available on the target platform, specially for multi-arch/targets compilers).

The next/other problem is that gcc compiler distinguish between used/maximum feature set and optimization for target arch...

Do not know a perfect example but a quick search found e.g. [1] (did not check if it works for cross-compiling)...

But a simple solution is to implement one/some cmake option(s) to enable target specific flags/optimizations (easy to not to use and/or override with global/package specific compiler flags)...

[1] https://github.com/seredat/karbowanec/blob/master/CMakeLists.txt

davidplowman: Would it work for you to have a TARGET variable which can default to (say) "armv8" but such that you can invoke cmake with "-DTARGET=other" or something like that?

pseiderer: Yes, should definitely work (would name it a little more specific like ENABLE_COMPILE_FLAGS_FOR_TARGET_ARCH?)...

davidplowman: Would this do the trick for you?
```
IF (NOT ENABLE_COMPILE_FLAGS_FOR_TARGET)
  set(ENABLE_COMPILE_FLAGS_FOR_TARGET "armv8")
endif()
if ("${ENABLE_COMPILE_FLAGS_FOR_TARGET}" STREQUAL "armv8")
  add_definitions(-mfpu=neon-fp-armv8 -ftree-vectorize)
endif()
```
If you were ok to put the changes we've discussed into your branch then I'd be happy to test and merge if it all works fine for our "default Pi user". Thanks!

pseiderer: Yes, works for the buildroot case, merge request updated...

PR #8, opened 2021-02-13
Fix option parser to use "analoggain" correctly.

naushir:
The option parser sets a default value of 0 for "gain" and "analoggain"
if not provided.  Since they both reference the same underlying variable,
if "analoggain" is set in the command line, the variable would be rest to
the default (0) since "gain" is lower in the option list.

### DIFF
--- a/options.hpp
+++ b/options.hpp
@@ -48,7 +48,7 @@ struct Options
 			 "Set a fixed shutter speed")
 			("analoggain", value<float>(&gain)->default_value(0),
 			 "Set a fixed gain value (synonym for 'gain' option)")
-			("gain", value<float>(&gain)->default_value(0),
+			("gain", value<float>(&gain),
 			 "Set a fixed gain value")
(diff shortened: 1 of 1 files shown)


naushir:
On a somewhat related note, should default for framerate be set to 0?  Right now it is set to 30, and as a consequence, libcamera-still has a preview rate of 30fps fixed.  I would have thought we would want variable fps here, so the default should be 0?

@davidplowman if you think this is better, I can make the change and package it in this PR.

davidplowman:
Hi Naush, thanks for fixing the analoggain thing, I always type the shorter "--gain" so never noticed!

Re. the preview framerate. I've been trying this and I think I still get the occasional obvious "wink". It's not like it was before (i.e. oscillating the whole time), it really happens only randomly and very occasionally. It feels like random timing "bad luck" or something, so maybe we need to have a look at that first.

In the meantime I'll apply this change.

PR #9, opened 2021-02-21
Display frame information in the titlebar

naushir:


naushir:
@davidplowman 

I'm sure there are better ways to allow the user to select what to display on the titlebar, so we don't have to take the second patch if you think it's overkill.

davidplowman:
Hi Naush, thanks for this feature, I really like it! I did want to have a few little meta-discussions about some bits of it...

1. I think we're measuring the framerate of the frames that actually get displayed, which is inherently limited by the display refresh rate. I wonder if we should put that measurement on the event loop side of things, rather than in the asynchronous preview thread, so that we could (in principle anyway) measure higher framerates? Just as an example, we could even measure the instantaneous framerate as part of the request completion callback, and put it in the message payload that we queue up, though I don't know if that's the best way.

2. GetFrameInfo seems to have sprouted quite a lot of arguments. Possibly I should have made a simple "FrameInfo" structure for it straight away, but I was evidently a bit lazy. Maybe now would be the time? A FrameInfo object might have a few simple methods, such as a constructor which makes it out of the ControlList metadata, perhaps a method to give its string representation? Though that depends a bit on how we handle other things...

3. If we're going to be passing around BufferMaps and their ControlList metadata together more of the time, should we pass them in a structure? I think there already is one - the RequestCompletePayload or something? - though it could have a less hideous name...

What do you think? Though I'm also aware this would start to make it all a more substantial change, which probably wasn't intended...

naushir:
> 1. I think we're measuring the framerate of the frames that actually get displayed, which is inherently limited by the display refresh rate. I wonder if we should put that measurement on the event loop side of things, rather than in the asynchronous preview thread, so that we could (in principle anyway) measure higher framerates? Just as an example, we could even measure the instantaneous framerate as part of the request completion callback, and put it in the message payload that we queue up, though I don't know if that's the best way.

Yes, that does seem reasonable.  However, it is a bit awkward as this framerate measuring code must be duplicated for each app's event loop, or have I got that wrong?

> 2. GetFrameInfo seems to have sprouted quite a lot of arguments. Possibly I should have made a simple "FrameInfo" structure for it straight away, but I was evidently a bit lazy. Maybe now would be the time? A FrameInfo object might have a few simple methods, such as a constructor which makes it out of the ControlList metadata, perhaps a method to give its string representation? Though that depends a bit on how we handle other things...

I did intend to add exactly such a class, but opted for the lazy approach instead... :)

> 3. If we're going to be passing around BufferMaps and their ControlList metadata together more of the time, should we pass them in a structure? I think there already is one - the RequestCompletePayload or something? - though it could have a less hideous name...
> 

I think I will add a FrameInfo class that populates from the ControlList.  This gets passed back though the RequestCompletePayload message.  Does it need to be an explicit struct holding the two?  Maybe a std::pair will suffice?

davidplowman:
> > 1. I think we're measuring the framerate of the frames that actually get displayed, which is inherently limited by the display refresh rate. I wonder if we should put that measurement on the event loop side of things, rather than in the asynchronous preview thread, so that we could (in principle anyway) measure higher framerates? Just as an example, we could even measure the instantaneous framerate as part of the request completion callback, and put it in the message payload that we queue up, though I don't know if that's the best way.
> 
> Yes, that does seem reasonable. However, it is a bit awkward as this framerate measuring code must be duplicated for each app's event loop, or have I got that wrong?

I was wondering about calculating the instantaneous framerate in the requestComplete method, and then putting the result in the RequestCompletePayload, that way everyone would get it, and we'd be calculating it a bit nearer the correct place. As a general thing, this is then a place to put any extra information that we decide we like, over and above what the ControlList allows.

> 
> > 1. GetFrameInfo seems to have sprouted quite a lot of arguments. Possibly I should have made a simple "FrameInfo" structure for it straight away, but I was evidently a bit lazy. Maybe now would be the time? A FrameInfo object might have a few simple methods, such as a constructor which makes it out of the ControlList metadata, perhaps a method to give its string representation? Though that depends a bit on how we handle other things...
> 
> I did intend to add exactly such a class, but opted for the lazy approach instead... :)
> 
> > 1. If we're going to be passing around BufferMaps and their ControlList metadata together more of the time, should we pass them in a structure? I think there already is one - the RequestCompletePayload or something? - though it could have a less hideous name...
> 
> I think I will add a FrameInfo class that populates from the ControlList. This gets passed back though the RequestCompletePayload message. Does it need to be an explicit struct holding the two? Maybe a std::pair will suffice?

I'm a bit in several minds about whether we pass back the metadata ControlList (as we do now), but with some extra values in the RequestCompletePayload (like the framerate). Or whether we turn that entire ControlList into a FrameInfo and return only that in the RequestCompletePayload. The latter is perhaps tidier, though it means we have plumbing to do whenever there's a new libcamera control that we decide we want to pass through. Maybe that's ok?

I wonder then about passing the RequestCompletePayload around (though with a better name) - I think I'd rather have meaningfully named fields. But I appreciate this is all becoming a bigger change. I do vaguely foresee a future (possibly this summer already if we get some interns) where we have "post-processors" that want to add their own metadata (such as neural network results), and such a structure would then be a place to put this... But as you can see, I'm a bit all over the place with this. Help is appreciated!!

naushir:
Now rebased on top of tree.

naushir:
@davidplowman, feel free to review when you can :)

egl_preview.hpp
```diff
@@ -56,4 +58,5 @@ class EglPreview : public Preview
 	int y_;
 	int width_;
 	int height_;
+	std::string infoText;
```
davidplowman, line 61: At the risk of being really annoying, libcamera-apps are generally using snake_case (just when we've started getting used to camelCase for libcamera itself), and members of (non-trivial) classes get a _ on the end...

libcamera_app.hpp
```diff
@@ -743,8 +744,21 @@ class LibcameraApp
...
+	{
+		const libcamera::FrameMetadata &metadata = buffer->metadata();
+		float exposure, analogue_gain, digital_gain, colour_gains[2];
+		char s[128];
+
+		GetFrameInfo(controls, &exposure, &analogue_gain, &digital_gain, colour_gains);
+		snprintf(s, sizeof(s), "Frame #%u: Exp %.2fms AnaGain %.2fx DigGain %.2fx RedGain %.2fx BlueGain %.2fx",
+				 metadata.sequence, exposure/1000.0, analogue_gain, digital_gain, colour_gains[0], colour_gains[1]);
+		return std::string(s);
+	}
```
davidplowman, line 762: I think it would be nicer to put the sequence number in the CompletedRequest, as we did the framerate. Then it would also be convenient just to pass the CompletedRequest in here as it doesn't need anything else?

The next commit possibly re-arranges some of this stuff, but could this member function be static?

preview.hpp
```diff
@@ -24,6 +25,7 @@ class Preview
 	{
 		done_callback_ = callback;
 	}
+	virtual void SetInfoText(const std::string &text) = 0;
```
davidplowman, line 28: I fine either way with this one, but given that 2 classes define an empty implementation for this method, maybe we should just supply that by default? Don't really mind though.

frame_info.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+        unsigned int sequence;
+        float exposure_time;
+        float analogue_gain;
+        float digital_gain;
+        std::array<float, 2> colour_gains;
+        float focus;
+        float fps;
+        bool aelock;
+};
```
davidplowman, line 47: As per comments for the previous commit, I think it might be neater to put the sequence number in the CompletedRequest, and just to pass that in to the constructor.

libcamera_app.hpp
```diff
@@ -744,20 +715,50 @@ class LibcameraApp
...
+				else if (t == "%focus")
+					value << frame_info.focus;
+				else if (t == "%aelock")
+					value << frame_info.aelock;
+
+				info_string.replace(pos, t.length(), value.str());
+			}
+		}
+		return info_string;
 	}
```
davidplowman, line 762: Looking at this 2nd commit, maybe we could make this a member of the FrameInfo? The LibcameraApp class already has too much in it...!

egl_preview.hpp
naushir: Oops, ill fix that.

libcamera_app.hpp
naushir: Sure. Will fix.

preview.hpp
naushir: Yes, makes sense

libcamera_app.hpp
naushir: I can add a `to_string()` method that takes in the token string and returns out the parsed string.

davidplowman: Except I think method names are CamelCase (or camelCase if private...). So much for being consistent!!

naushir:
@davidplowman, all fixed applied.  I've reworked some of the intermediate patches as well for clarity.

PR #10, opened 2021-02-22
Buffer and metadata rework

davidplowman:
@naushir
Here's that re-work of the buffer and metadata passing that we were talking about. Needless to say, the original plan bit the dust almost immediately on contact with reality - I found I was making quite a lot of use of ControlLists, and didn't really want to rewrite that. The aim was never really to produce de-libcamera-ified apps, so I backed off a bit. Instead I have:

1. Renamed the RequestCompletePayload object as a BufferMetadataSet. Not sure I love this name either, but it's not as bad. Still open to other suggestions, though!

2. I've added the instantaneous framerate to the BufferMetadataSet, and it gets calculated in the requestComplete callback.

3. Now we pass BufferMetadataSets round everywhere, rather than BufferMaps. This way, the metadata ControlList and the framerate are always available.

So not as much of a change as originally intended, but nonetheless I think it should make adding things like the smart-title-bar feature more convenient. See what you think...

### DIFF
--- a/libcamera_app.hpp
+++ b/libcamera_app.hpp
@@ -57,9 +57,9 @@ public:
 	using BufferMap = Request::BufferMap;
 	using Size = libcamera::Size;
 	using Rectangle = libcamera::Rectangle;
-	struct RequestCompletePayload
+	struct CompletedRequest
 	{
-		RequestCompletePayload(BufferMap const &b, ControlList const &m)
+		CompletedRequest(BufferMap const &b, ControlList const &m)
 			: buffers(b), metadata(m) {}
 		BufferMap buffers;
 		ControlList metadata;
@@ -70,7 +70,7 @@ public:
 		RequestComplete,
 		Quit
 	};
-	typedef std::variant<RequestCompletePayload, QuitPayload> MsgPayload;
+	typedef std::variant<CompletedRequest, QuitPayload> MsgPayload;
 	struct Msg
 	{
 		Msg(MsgType const &t, MsgPayload const &p) : type(t), payload(p) {}
@@ -673,7 +673,7 @@ private:
 		if (request->status() == Request::RequestCancelled)
 			return;
 
-		RequestCompletePayload payload(request->buffers(), request->metadata());
+		CompletedRequest payload(request->buffers(), request->metadata());
 		{
 			request->reuse();
 			std::lock_guard<std::mutex> lock(free_requests_mutex_);
--- a/libcamera_app.hpp
+++ b/libcamera_app.hpp
@@ -63,6 +63,7 @@ public:
 			: buffers(b), metadata(m) {}
 		BufferMap buffers;
 		ControlList metadata;
+		float framerate;
 	};
 	struct QuitPayload {};
 	enum class MsgType
@@ -375,6 +376,7 @@ public:
 			throw std::runtime_error("failed to start camera");
 		controls_.clear();
 		camera_started_ = true;
+		last_timestamp_ = 0;
 
 		camera_->requestCompleted.connect(this, &LibcameraApp::requestComplete);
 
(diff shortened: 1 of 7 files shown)


naushir:
Not too keen on BufferMetadataSet either :-)

How about CompletedFrame, CompletedRequest, or just Frame?

naushir:
I did like the idea of a `FrameInfo` class that presents the ControlList in a nice easy way for us to use though.  It would have been much nicer than the existing `GetFrameInfo()` function.  Perhaps we can still do something with it?  Here is a sketch of what I was thinking:

```
struct FrameInfo
{
    FrameInfo(ControlList *ctrls, unsigned int _sequence)
        : sequence(_sequence), exposure_time(0.0), digital_gain(0.0), colour_gain({0.0,0.0}),
          focus(0.0), aelock(false)
    {
			if (ctrls.contains(controls::ExposureTime))
				exposure_time = ctrls.get<int32_t>(controls::ExposureTime);

			if (ctrls.contains(controls::AnalogueGain))
				analogue_gain = ctrls.get(controls::AnalogueGain);

			if (ctrls.contains(controls::DigitalGain))
				digital_gain = ctrls.get(controls::DigitalGain);

			if (ctrls.contains(controls::DigitalGain))
			{
				libcamera::Span<const float> gains = ctrls.get(controls::ColourGains);
				colour_gains[0] = gains[0], colour_gains[1] = gains[1];
			}

			if (ctrls.contains(controls::FocusFoM))
				focus = ctrls.get(controls::FocusFoM);

    		if (ctrls.contains(controls::AeLocked))
				aelock = ctrls.get(controls::AeLocked);
    }

    unsigned int sequence;
	float exposure;
    float analogue_gain;
    float digital_gain;
    std::array<float, 2> colour_gains[2];
    float focus;
	float fps;
    bool aelock;
}
```

libcamera_app.hpp
```diff
@@ -785,4 +795,6 @@ class LibcameraApp
 	// For setting camera controls.
 	std::mutex control_mutex_;
 	ControlList controls_;
+	// Other:
+	uint64_t last_timestamp_ = 0;
 };
```
naushir, line 804: `last_timestamp_`  probably wants to be initialised to in `StartCamera()` instead of here, as start/stop/start cycles will not use stale values.

davidplowman:
Thanks, Naush. I could probably go with "CompletedRequest". Anything with "Frame" sounds a bit like it's a single frame buffer. I want to convey that this is "everything", all the frame buffers, metadata etc.

I agree that a FrameInfo is a good idea and I think we should do it. I'm just not sure I want to put it in the CompletedRequest. As I said previously, there's stuff that would have to be re-written, there's more that would have to go in it (colour matrices, for example), and any time libcamera added a control we'd have to go and do some more plumbing. But I think we should define the class, allow it to be constructed from a CompletedRequest, and then it provides a convenient way for people to get hold of all that information. And it could have a toString method!

I'll go and patch stuff up and re-submit an updated version.

libcamera_app.hpp
```diff
@@ -445,7 +451,7 @@ class LibcameraApp
 		StreamDimensions(video_stream_, w, h, stride);
 		return video_stream_;
 	}
-	void QueueRequest(BufferMap const &buffers)
+	void QueueRequest(CompletedRequest const &buffer_set)
 	{
```
naushir, line 455: Sorry, @davidplowman, but would you consider renaming `buffer_set` to `completed_request` to be consistent with the new name?

With or without that change, I think this is good to merge.

davidplowman: Sure, I did wonder a bit about that... let me fix it up and then we can hit the big green button.

naushir:
Looks good, ill go ahead and merge it.

PR #11, opened 2021-02-24
Fix some compiler warning.

naushir:
- Unused return value from getline().
- printf format specifiers for 64-bit integers.
- Warning about narrowing conversion from int to bool.

### DIFF
--- a/output.cpp
+++ b/output.cpp
@@ -5,6 +5,7 @@
  * output.cpp - video stream output base class
  */
 
+#include <cinttypes>
 #include <stdexcept>
 
 #include "output.hpp"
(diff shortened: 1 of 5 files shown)


naushir:
@davidplowman

davidplowman:
Thanks for these, Naush. I'll commit them in just a moment. Any idea why my compiler (gcc 8.3) hasn't flagged them?

naushir:
Not sure... I'm on gcc 9.3, so it's probably a version thing.  Also, I am compiling on x64/Linux.

davidplowman:
Ah yes. Well it's probably a good thing to be using different compilers. Perhaps I should switch to clang...!

output.cpp
```diff
@@ -60,7 +61,7 @@ void Output::OutputReady(void *mem, size_t size, int64_t timestamp_us, bool keyf
 
 	// Save timestamps to a file, if that was requested.
 	if (fp_timestamps_)
-		fprintf(fp_timestamps_, "%lld.%03lld\n", last_timestamp_ / 1000, last_timestamp_ % 1000);
+		fprintf(fp_timestamps_, PRId64 ".%03" PRId64 "\n", last_timestamp_ / 1000, last_timestamp_ % 1000);
```
marv, line 64: This is missing a `"%"` before the first `PRId64`, no?

naushir: Yes, good catch!

@davidplowman can you force push this fix. Sorry about missing that.

davidplowman: Will do. I might add something for this to the "smoke tests", that is, test.py. In general I'm keen to keep beefing those up as we go...

PR #12, opened 2021-02-24
Add a test to check timestamp files written by libcamera-vid

davidplowman:
We perform a very simple check that the file exists, that the header
is as we expect, and that the first two entries are numeric and
increasing.


davidplowman:
@naushir
Do you want to check this over and merge it once you're happy there are no obvious clangers? Thanks!

PR #21, opened 2021-03-22
libcamera-raw: Fix video capture with full res raw images

davidplowman:
The number of raw buffers must be set up to match the number of output
buffers, unconditionally. libcamera_app.hpp requires all streams that
will feature in the same requests to have the same number of buffers
available.


davidplowman:
@naushir Hi Naush, another reasonably trivial one for you, I'm hoping...!

PR #24, opened 2021-04-09
mjpeg_encoder: fix bug that truncated jpegs to 4096 bytes

davidplowman:
Fixing up some type warnings had caused the jpeg length not to be
returned correctly from the encodeJPEG method.

Test script has also been altered to spot this.

Fixes: b20dc097621a ("fix buildroot crosscompile (#1)")


davidplowman:
@naushir Should have spotted this at the time.

PR #28, opened 2021-04-20
Fix deadlock in H.264 encoder

naushir:
There was a case where two mutexes were acquired in different orders
causing a deadlock.


naushir:
@davidplowman

PR #29, opened 2021-04-21
Make interface from libcameraApp to Encoder cleaner

davidplowman:
Previously LibcameraEncoder::EncodeBuffer used to return a handle
that would in theory allow an encoder to return buffers out-of-order
(though LibcameraEncoder::encodeBufferDone never handled this). This
entangles the LibcameraEncoder and the Encoder together more than is
necessary.

Instead, use the memory pointer, which we already have, as the
handle. This will make it easier to handle out-of-order completion,
should we choose to do so in future.

### DIFF
--- a/encoder.hpp
+++ b/encoder.hpp
@@ -11,7 +11,7 @@
 
 #include "video_options.hpp"
 
-typedef std::function<void(int)> InputDoneCallback;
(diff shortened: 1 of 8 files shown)


davidplowman:
@naushir 
I believe this improves the previous fix, meaning we could support out-of-order buffer completion should we want to.

PR #35, opened 2021-05-17
Auto-detect if Neon is available on this platform

davidplowman:
We need to omit the -mfpu=neon-fp-armv8 flag for Pis without Neon
support.


davidplowman:
@naushir Another one, I'm afraid. This patch resolves this documentation issue: https://github.com/raspberrypi/documentation/issues/1902

PR #38, opened 2021-06-01
Frame duration limits

davidplowman:
@naushir
Hi Naush - can you double-check and merge for me, please?


davidplowman:
This is correct now!

PR #43, opened 2021-06-29
encoder: Do not explicitly set buffer size and bytesperline in format struct

naushir:
Given that we are using dmabufs, these fields do not seem to be necessary to
fill in the format structure.


naushir:
@davidplowman this does not seem to break anything!

davidplowman:
Well that's much better then!

PR #44, opened 2021-06-29
Build fixups for latest libcamera version

davidplowman:
libcamera is now the "libcamera" (rather than just "camera") package,
and has also had a "base" library split out which we must now link.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,7 +41,7 @@ add_library(encoders encoder.cpp null_encoder.cpp h264_encoder.cpp mjpeg_encoder
 project(libcamera-still)
 
 find_package(PkgConfig REQUIRED)
-pkg_check_modules(CAMERA REQUIRED camera)
+pkg_check_modules(CAMERA REQUIRED libcamera)
 pkg_check_modules(LIBDRM REQUIRED libdrm)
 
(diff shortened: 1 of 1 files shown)


davidplowman:
@naushir 
This should enable libcamera-apps to build and link correctly against the newly restructured libcamera.

naushir:
Looks good.

CMakeLists.txt
```diff
@@ -53,7 +53,10 @@ find_library(JPEG_LIBRARY jpeg REQUIRED)
 find_library(TIFF_LIBRARY tiff REQUIRED)
 find_library(PNG_LIBRARY png REQUIRED)
 find_library(LIBCAMERA_LIBRARY libcamera.so REQUIRED)
-message(STATUS "${LIBCAMERA_LIBRARY}")
+find_library(LIBCAMERA_BASE_LIBRARY libcamera-base.so REQUIRED)
```
kbingham, line 56: Does find_library use pkg-config? This should be provided by the libcamera.pc pkg-config so you shouldn't have to specify it manually....

If you do - it means I failed :-(

davidplowman: I daresay there's a "proper" way to do it, but I claim absolutely no expertise in cmake. I'm open to offers, of course...

PR #45, opened 2021-07-02
Build 32-bit binaries by default that run on all Pis.

davidplowman:
Only apply Pi 3/4 specific optimisations if explicitly reuested to do
so on the command line. This ensures that by default (in 32-bit
Raspberry Pi OS distributions) the executables will work on all
devices.


davidplowman:
@naushir 
Hi Naush, so the purpose of this is to make it possible to distribute these apps as part of our regular 32-bit OS. By default, we'll build binaries that run on all Pis, of whatever flavour. Folks who want Neon optimisations for Pi 3/4 can request them easily if they build the code for themselves. On our 64-bit OS, we can request Neon optimisations regardless.

PR #48, opened 2021-07-06
Updated formatting and style checker tool

naushir:
This set of commits updates all source files to have a consistent format/style via clang-format.
We also add a checkstyle.py script that can be used as a git hook to ensure all commits match our coding style.


naushir:
@davidplowman

PR #49, opened 2021-07-07
checkstyle: Update license to "GPL-2.0-or-later" for the script

naushir:
Use the original authorship licence of the checkstyle.py script file. It was
erroneously set to the default of "BSD-2-Clause" for this repository.

### DIFF
--- a/checkstyle.py
+++ b/checkstyle.py
@@ -1,6 +1,6 @@
 #!/usr/bin/python3
 
-# SPDX-License-Identifier: BSD-2-Clause
+# SPDX-License-Identifier: GPL-2.0-or-later
 #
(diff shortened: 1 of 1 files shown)


naushir:
@davidplowman and @kbingham
Here is a fix-up to the checkstyle.py license.

davidplowman:
I don't know where this file originated, but do we need to include anyone else's copyright here?

kbingham:
> I don't know where this file originated, but do we need to include anyone else's copyright here?

No, Laurent is the original author of this.

naushir:
I think the only thing missing is the original file's license.

davidplowman:
OK, well I'll push the big green button in an hour or so if no one has any further objections!

PR #50, opened 2021-07-07
Remove c++17 reference from .clang-format

davidplowman:
The Pi repos are not yet shipping with a version of clang-format that
supports c++17, so this parameter causes commits to fail if you are
doing them directly from the Pi.

For the time being little is lost by omitting it, and we can replace
it once the repositories have moved on.


davidplowman:
@naushir 
HI Naush - another little one for you!

PR #51, opened 2021-07-08
Include tidy ups

davidplowman:


davidplowman:
@naushir 
Hi Naush, yes - yet more stuff! This set is almost exclusively just tidy-ups to the includes, though the final commit fixes a little test timing niggle when I test with smaller sensors (e.g. ov5647). It's all been made possible by creating that separate libcamera_app.cpp - which shows what a good idea that was!

I think the include improvements are quite significant. The first commit is the biggest because it gets rid of that whole X11 "#define Status int" thing which breaks some of the libcamera includes. Then apart from that it's just a case of moving a few header files out of libcamera_app.hpp, and the derived xxx_preview.hpp files can now (because no one else needs to include them) disappear into the corresponding xxx_preview.cpp files.

Anyway, there are no functional changes, everything works exactly the same. But no urgency seeing as I'm away for a bit!

Thanks!

libcamera_app.cpp
```diff
@@ -40,27 +43,27 @@ void LibcameraApp::OpenCamera()
...
 					std::cout << "Made DRM preview window" << std::endl;
 			}
 			catch (std::exception const &e)
 			{
 				std::cout << "Preview window unavailable" << std::endl;
-				preview_ = std::make_unique<NullPreview>(options_.get());
+				preview_ = std::unique_ptr<Preview>(make_null_preview(options_.get()));
 			}
 		}
 	}
```
naushir, line 69: I'm curious why we added make_xxx_preview() functions over the existing std::make_unique mechanism.  Is it that we have no reference to xxxPrieview class definitions now, or am I missing something else?

davidplowman: Yes, it's so that libcamera_app.cpp doesn't have to drag in constructors for NullPreview, EglPreview etc.

naushir: Got it!

PR #52, opened 2021-07-08
Reorganise the libcamera-apps source tree.

naushir:
The source tree is becoming a bit crowded to use a flat structure now. This
commit reorganises the tree into the following directories:

apps - Top level applications
encoder - Various stream encoders
image - Still image helpers
output - Output network/file sink helpers
preview - EGL/DRM and null preview helpers


naushir:
@davidplowman could you please test this on your end to make sure I've not missed anything?

davidplowman:
All builds and works fine. Just a couple of things:

I think the hpp files in preview are all redundant now (I think a previous commit of mine removed them, though that may have gone wrong).

I was wondering whether libcamera_encoder.hpp belongs in core rather than apps as it doesn't actually generate an application file, and it's kind of "core" functionality to anyone who wants to encode stuff. But I might be splitting hairs there!

Otherwise looking good though shall continue to scrutinise for a bit longer...

davidplowman:
Sorry, I meant preview.hpp should obviously stay, just not the others!

davidplowman:
Ah, another thing I've noticed. It seems to make a library out of every cpp file, for example there's a "null_preview" library, an "egl_preview" library and a "drm_preview" library. Could we maybe just put those objects into a single "preview" library? And obviously corresponding changes in the other folders. I think that would be a bit neater. What do you think?

naushir:
> Sorry, I meant preview.hpp should obviously stay, just not the others!

Sure I will remove it.  This might have been a rebase error on my part.

naushir:
> Ah, another thing I've noticed. It seems to make a library out of every cpp file, for example there's a "null_preview" library, an "egl_preview" library and a "drm_preview" library. Could we maybe just put those objects into a single "preview" library? And obviously corresponding changes in the other folders. I think that would be a bit neater. What do you think?

I thought this did not happen, as I define an ``INTERFACE`` library consisting of all three preview libs, and that gets linked with the libcamera_app core library.  Do you think this is not happening?

naushir:
Barring anything major, ill do all the fixups tomorrow and merge this change.

preview/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+project(drm_preview)
+add_library(drm_preview drm_preview.cpp)
+target_link_libraries(drm_preview ${DRM_LIBRARY})
+
+project(null_preview)
+add_library(null_preview null_preview.cpp)
+
+add_library(preview INTERFACE)
+target_link_libraries(preview INTERFACE egl_preview drm_preview null_preview)
+
```
naushir, line 27: This is the bit that makes the INTERFACE library.

core/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+cmake_minimum_required(VERSION 3.6)
+
+find_package(Boost REQUIRED COMPONENTS program_options)
+
+add_library(libcamera_app libcamera_app.cpp)
+set_target_properties(libcamera_app PROPERTIES PREFIX "" IMPORT_PREFIX "")
+target_link_libraries(libcamera_app pthread images preview ${LIBCAMERA_LIBRARIES} ${Boost_LIBRARIES})
+
+install(TARGETS libcamera_app LIBRARY DESTINATION lib ARCHIVE DESTINATION lib)
+
```
naushir, line 10: This links the preview INTERFACE library.  Perhaps this is not doing what I think it is meant to do?

davidplowman:
> > Ah, another thing I've noticed. It seems to make a library out of every cpp file, for example there's a "null_preview" library, an "egl_preview" library and a "drm_preview" library. Could we maybe just put those objects into a single "preview" library? And obviously corresponding changes in the other folders. I think that would be a bit neater. What do you think?
> 
> I thought this did not happen, as I define an `INTERFACE` library consisting of all three preview libs, and that gets linked with the libcamera_app core library. Do you think this is not happening?

The make output looks a bit surprising to me. Checking afterwards I see:
`
pi@pi4-davidp:~/libcamera-apps/build $ find . -name "*.a"
./preview/libnull_preview.a
./preview/libdrm_preview.a
./preview/libegl_preview.a
./output/liboutputs.a
./encoder/libencoders.a
./image/libdng_base.a
./image/libimages_base.a
./image/libjpeg_base.a
./image/libpng_base.a
./core/libcamera_app.a
`
some of which I was expecting (liboutputs.a, for example) but there seem to be a few I wasn't (libnull_preview.a but no libpreview.a). Also any reason why some of them have _base on the end? Is that to avoid name confliicts?

davidplowman:
Yes, seeing as I'm off it's probably best if you merge it yourself once you're ready. I think I've found everything that was bothering me so once those are good... :)

naushir:
> > > Ah, another thing I've noticed. It seems to make a library out of every cpp file, for example there's a "null_preview" library, an "egl_preview" library and a "drm_preview" library. Could we maybe just put those objects into a single "preview" library? And obviously corresponding changes in the other folders. I think that would be a bit neater. What do you think?
> > 
> > 
> > I thought this did not happen, as I define an `INTERFACE` library consisting of all three preview libs, and that gets linked with the libcamera_app core library. Do you think this is not happening?
> 
> The make output looks a bit surprising to me. Checking afterwards I see:
> 
> ```
> pi@pi4-davidp:~/libcamera-apps/build $ find . -name "*.a"
> ./preview/libnull_preview.a
> ./preview/libdrm_preview.a
> ./preview/libegl_preview.a
> ./output/liboutputs.a
> ./encoder/libencoders.a
> ./image/libdng_base.a
> ./image/libimages_base.a
> ./image/libjpeg_base.a
> ./image/libpng_base.a
> ./core/libcamera_app.a
> ```
> 
> some of which I was expecting (liboutputs.a, for example) but there seem to be a few I wasn't (libnull_preview.a but no libpreview.a). Also any reason why some of them have _base on the end? Is that to avoid name confliicts?

Might have to dig out the old cmake manual to see what is going on!

PR #53, opened 2021-07-09
cmake: Use pkg_check_modules for libcamera, x11, epoxy, and libdrm li…

naushir:
…braries

There was a mix of pkg_check_modules and find_library usage that is now cleaned
up.


kbingham:
Aha excellent - this resolves my worries about having to specify libcamera-base explicitly.

PR #57, opened 2021-07-19
Support files larger than 2GB

davidplowman:
This define switches us to large files on 32 bit systems, and is
harmless on 64 bit systems (where large files are the default).


davidplowman:
@naushir Having hunted around a bit this seemed like maybe it was the preferred way to achieve this...!

davidplowman:
Ah, I seem to have added an unrelated second commit to the same pull request. But they're both tiny so I think we can manage!

PR #58, opened 2021-07-22
build: Fix error linking to libcamera-base.so

naushir:
It's a bit strange why this did not fail all the time.
I suspect the "wrong" variable may have been cached by cmake, so it kept working when it should not have been!

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -4,7 +4,7 @@ find_package(Boost REQUIRED COMPONENTS program_options)
 
 add_library(libcamera_app libcamera_app.cpp)
 set_target_properties(libcamera_app PROPERTIES PREFIX "" IMPORT_PREFIX "")
-target_link_libraries(libcamera_app pthread images preview ${LIBCAMERA_LIBRARIES} ${Boost_LIBRARIES})
+target_link_libraries(libcamera_app pthread images preview ${LIBCAMERA_LINK_LIBRARIES} ${Boost_LIBRARIES})
 
 install(TARGETS libcamera_app LIBRARY DESTINATION lib ARCHIVE DESTINATION lib)
 


naushir:
@davidplowman

davidplowman:
I'm not really understanding this either. As far as I can see the values are:

LIBCAMERA_LIBRARIES: camera;camera-base
LIBCAMERA_LINK_LIBRARIES: /usr/local/lib/arm-linux-gnueabihf/libcamera.so;/usr/local/lib/arm-linux-gnueabihf/libcamera-base.so

so the difference is that the second explicitly gives the full name - but they're in the "standard" place and it seems they should be found anyway. Certainly if I delete my build directory and rebuild from scratch, everything seems to work as it stands...

davidplowman:
@kbingham Wondering if you had any ideas here. One of our users has had the problem whereby our executables link, but when run it can't find libcamera-base.so and aborts (though it does find libcamera.so). But if we run ```sudo ldconfig``` the problems goes away. Any thoughts?

davidplowman:
I've done another full-install-from-scratch and everything works fine. I wonder if there's some issue whereby a pre-libcamera-base version of libcamera doesn't quite upgrade the ld.so.cache appropriately when that extra library appears?

davidplowman:
OK, so I've been through the whole process again starting with a clean image. This time I checked out libcamera and libcamera-apps from early June. Everything builds and runs. I notice that /etc/ld.so.cache contains only libcamera.so, as expected.

Now I update libcamera to top of tree, rebuild and re-install. libcamera-base.so has now appeared, but I note ld.so.cache still only refers to libcamera.so. Anyway, I now update and build libcamera-apps (no errors), and sure enough, libcamera-base.so is not found when I try to run the executable. Running ```sudo ldconfig``` makes libcamera-base.so appear in ld.so.cache, and all is now well.

So the workaround for now does indeed seem to be ```sudo ldconfig```. I've filed a bug at https://bugs.libcamera.org/show_bug.cgi?id=68

davidplowman:
After a bit of delving around it seems a bit unclear whose responsibility it is to update the ld.so.cache. Under the circumstances it's probably worthwhile making this change as it solves the problem for users without having to find the ```sudo ldconfig``` workaround. and would appear to have no bad side-effects.

kbingham:
> @kbingham Wondering if you had any ideas here. One of our users has had the problem whereby our executables link, but when run it can't find libcamera-base.so and aborts (though it does find libcamera.so). But if we run `sudo ldconfig` the problems goes away. Any thoughts?

Hi David, sorry for the delayed reply - I was on holiday last week so I've seen this too late.
I'm not particularly proficient in cmake, so I'm not sure about what effect the specific change has - but if it works it sounds helpful ;-)

I see from other discussions it may also be due to ldconfig not running after package installations - so it does sound a bit more painful, and I wouldn't want users to manually run sudo ldconfig after installations. Perhaps that's handled by the distributions when they install a package, and not expected to be done at 'ninja install' time (especially as the ninja install might be installing to a temporary location) ...

Anyway, sounds like this solves your problems for now.

PR #59, opened 2021-07-23
Add support for a low resolution stream

davidplowman:
Viewfinder and video modes can supply an extra "lores" stream,
specified by the --lores-width and --lores-height command line
parameters.

### DIFF
--- a/core/libcamera_app.cpp
+++ b/core/libcamera_app.cpp
@@ -122,7 +122,11 @@ void LibcameraApp::ConfigureViewfinder()
 	if (options_->verbose)
 		std::cout << "Configuring viewfinder..." << std::endl;
 
-	configuration_ = camera_->generateConfiguration({ StreamRole::Viewfinder });
+	bool have_lores_stream = options_->lores_width && options_->lores_height;
+	StreamRoles stream_roles = { StreamRole::Viewfinder };
+	if (have_lores_stream)
+		stream_roles.push_back(StreamRole::Viewfinder);
+	configuration_ = camera_->generateConfiguration(stream_roles);
 	if (!configuration_)
 		throw std::runtime_error("failed to generate viewfinder configuration");
 
@@ -147,12 +151,26 @@ void LibcameraApp::ConfigureViewfinder()
 	// Now we get to override any of the default settings from the options_->
 	configuration_->at(0).pixelFormat = libcamera::formats::YUV420;
 	configuration_->at(0).size = size;
+
+	if (have_lores_stream)
+	{
+		Size lores_size(options_->lores_width, options_->lores_height);
+		lores_size.alignDownTo(2, 2);
+		if (lores_size.width > size.width || lores_size.height > size.height)
+			throw std::runtime_error("Low res image larger than viewfinder");
+		configuration_->at(1).pixelFormat = libcamera::formats::YUV420;
+		configuration_->at(1).size = lores_size;
+		configuration_->at(1).bufferCount = configuration_->at(0).bufferCount;
+	}
+
 	configuration_->transform = options_->transform;
 
 	configureDenoise(options_->denoise == "auto" ? "cdn_off" : options_->denoise);
 	setupCapture();
 
 	streams_["viewfinder"] = configuration_->at(0).stream();
+	if (have_lores_stream)
+		streams_["lores"] = configuration_->at(1).stream();
 
 	if (options_->verbose)
 		std::cout << "Viewfinder setup complete" << std::endl;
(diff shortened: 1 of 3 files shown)


davidplowman:
@naushir
Here's what I'm thinking of for creating a low resolution stream. Not entirely sure about all of it, or indeed if we even want to merge it yet, but something to think about. I've not added a low res stream capability when capturing stills - it doesn't really seem like something we'd need?

core/libcamera_app.hpp
```diff
@@ -194,6 +195,7 @@ class LibcameraApp
 	Stream *still_stream_ = nullptr;
 	Stream *raw_stream_ = nullptr;
 	Stream *video_stream_ = nullptr;
+	Stream *lores_stream_ = nullptr;
 	FrameBufferAllocator *allocator_ = nullptr;
```
naushir, line 199: I wonder if we should replace this a std::map<std::string, Stream *> to make things a bit easier?
Then all our accessor functions can also disappear.

core/libcamera_app.cpp
```diff
@@ -122,7 +122,11 @@ void LibcameraApp::ConfigureViewfinder()
 	if (options_->verbose)
 		std::cout << "Configuring viewfinder..." << std::endl;
 
-	configuration_ = camera_->generateConfiguration({ StreamRole::Viewfinder });
+	bool have_lores_stream = options_->lores_width && options_->lores_height;
+	StreamRoles stream_roles = { StreamRole::Viewfinder };
+	if (have_lores_stream)
+		stream_roles.push_back(StreamRole::Viewfinder);
+	configuration_ = camera_->generateConfiguration(stream_roles);
 	if (!configuration_)
```
naushir, line 130: Not sure I get this logic.  Why do a push_back on stream_roles conditionally?  Also, stream_roles will have 2 elements of StreamRole::Viewfinder.

core/libcamera_app.cpp
```diff
@@ -147,12 +151,25 @@ void LibcameraApp::ConfigureViewfinder()
...
 	configuration_->at(0).pixelFormat = libcamera::formats::YUV420;
 	configuration_->at(0).size = size;
+
+	if (have_lores_stream)
+	{
+		Size lores_size(options_->lores_width, options_->lores_height);
+		lores_size.alignDownTo(2, 2);
+		if (lores_size.width > size.width || lores_size.height > size.height)
+			throw std::runtime_error("Low res image larger than viewfinder");
+		configuration_->at(1).pixelFormat = configuration_->at(0).pixelFormat;
```
naushir, line 161: Do we need to be a bit careful here?  If configuration_->at(0).pixelFormat is an RGB format, the low res output will not be able to do that.  Although, I think the pipeline handler might correct this, not sure.

core/libcamera_app.hpp
davidplowman: Yes, I agree that maybe something like a map might be tidier now. Are you thinking of something like "app->Stream("still")" in place of "app->StillStream()", or of exposing the map itself? I think I still prefer "if (app->StillStream()) ..." slightly to "if (app->Streams().find("still") != app->Streams().end()) ..."

core/libcamera_app.cpp
davidplowman: I guess I didn't want to force the creation of the low res stream if it wasn't needed. Unless it gets created anyway. But maybe that depends on whether colour denoise is running, or what size lo res stream we ask for...?? Maybe I'm a bit confused!

On the other point, I assume there's no problem in having two streams with the "viewfinder" role? I'm not too sure what other "role" I would give it!

core/libcamera_app.cpp
davidplowman: Good point. I should hardcode the lo res stream always to YUV420.

core/libcamera_app.hpp
naushir: You could use if (app->Streams("still").count) as a shortcut, but I am not too fussed wither way.

core/libcamera_app.cpp
naushir: Makes sense, I did not think about it that way. I don't think there is any problem with having multiple Viewfinder roles in that case.

davidplowman:
Revised version. There are now 2 commits. The first re-organises the stream pointers a bit which does make the second commit (that adds the lores stream) a bit tidier, I think.

davidplowman:
I'm thinking that we shouldn't merge this here, I'd rather put this into the work on post-processing and then we'll pull it all in from there in due course. Still, the discussion has been helpful.

PR #61, opened 2021-07-29
Post-processing: Add annotate stage

davidplowman:
The annotate stage writes text onto the top of the image and supports
the same % directives as the FrameInfo class (hence also matching the
--info-text option). It is implemented using OpenCV.


davidplowman:
@naushir @RaspiShonak

post_processing_stages/annotate_cv_stage.cpp
```diff
@@ -0,0 +1,113 @@
...
+bool AnnotateCvStage::Process(CompletedRequest &completed_request)
+{
+	libcamera::Span<uint8_t> buffer = app_->Mmap(completed_request.buffers[stream_])[0];
+	FrameInfo info(completed_request.metadata);
+	info.sequence = completed_request.sequence;
+
+	// Other post-processing stages can supply metadata to update the text.
+	completed_request.post_process_metadata.Get("annotate.text", text_);
+	std::string text = info.ToString(text_);
+
```
naushir, line 89: How does "annotate.text" get populated to start with?  Seems to be missing from this change?

naushir (commented):
Looks good!

PR #62, opened 2021-07-30
Lores stream

davidplowman:
@naushir @RaspiShonak
There are 3 commits here:

- The first replaces the individual stream pointers in the app by a map (Naush suggested this a little while back, I think).
- The second adds a "lo res" stream to the viewfinder and video use cases.
- The final commit amends the face detect stage to use the low resolution stream so as to avoid the resize, but you do have to add something like `--lores-width 640 --lores-height 480` to the command line.


naushir:
Looks good to me.

davidplowman:
Shall we push the big green button, then?

PR #63, opened 2021-07-30
build: Use shared rather than static libraries

davidplowman:
A small tweak is required in post_processing_stage.cpp to ensure the
"stages" table is initialised before anyone tries to put anything in
it.


davidplowman:
@naushir Do we want to switch to shared libraries now?

naushir:
Looks good to me.  Presumably the make install stage does the right thing with copying the *.so files?

davidplowman:
Good point - I'll just try that.

Yes, it seems to "install" as expected and run OK (after `sudo ldconfig`, of course!).

PR #64, opened 2021-08-02
post-processing: negate: optimise for non-vectorising builds

davidplowman:
The loop can be better optimised when the vectorising compiler cannot
be used.

### DIFF
--- a/post_processing_stages/negate_stage.cpp
+++ b/post_processing_stages/negate_stage.cpp
@@ -43,11 +43,13 @@ void NegateStage::Configure()
 
 bool NegateStage::Process(CompletedRequest &completed_request)
 {
-	int w, h, stride;
 	libcamera::Span<uint8_t> buffer = app_->Mmap(completed_request.buffers[stream_])[0];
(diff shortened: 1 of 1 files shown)


davidplowman:
@naushir 
For my amusement, I timed the loop for debug, release and vectorising builds, using either uint8_t, uint32_t or uint64_t operations. The results (in ms per 3MP frame, running single-threaded) were:
```
             uint8_t  uint32_t  uint64_t
Debug        900      230       120
Release      230      120       120
Vectorising  60       60        60
```
I've assumed we can rely on the buffers being 4-byte aligned. Do you think 8-byte alignment could be assumed, as that would run faster in a debug buidl?

naushir:
For YUV420 formats, they will be 8-byte aligned.  However, other formats will currently be 4-byte aligned.  There's nothing to stop us from changing all to 8-byte alignment.... but it sort of does not seem worth it for only improving Debug builds.

davidplowman:
Sounds godd as it is, then. Do you want to push the merge button?

PR #65, opened 2021-08-11
post-processing: Added ObjectClassifierTfStage

RaspiShonak:
Download tflite image classification model from this link: https://storage.googleapis.com/download.tensorflow.org/models/mobilenet_v1_2018_08_02/mobilenet_v1_1.0_224_quant.tgz

Extract all files

Download and extract labels file using:
curl https://storage.googleapis.com/download.tensorflow.org/models/mobilenet_v1_1.0_224_frozen.tgz  | tar xzv -C /tmp  mobilenet_v1_1.0_224/labels.txt

Add full file-path of mobilenet_v1_1.0_224_quant.tflite to the object_classify_tf_stage.json file for "model_name"

Add full file path of labels.txt to the object_classify_tf_stage.json file for "labels_name"

Signed-off-by: Shonak Joshi <shonak.joshi@raspberrypi.com>

### DIFF
--- a/core/post_processing_stage.hpp
+++ b/core/post_processing_stage.hpp
@@ -9,6 +9,7 @@
 
 #include <boost/property_tree/json_parser.hpp>
 #include <boost/property_tree/ptree.hpp>
+#include <chrono>
 
 namespace libcamera
 {
@@ -45,6 +46,19 @@ public:
 	virtual void Teardown();
 
 protected:
+	// Helper to calculate the execution time of any callable object and return it in as a std::chrono::duration.
+	// For functions returning a value, the simplest thing would be to wrap the call in a lambda and capture
+	// the return value.
+	template <class R = std::micro, class T = std::chrono::steady_clock, class F, class... Args>
+
+	static auto ExecutionTime(F &&f, Args &&... args)
+	{
+		auto t1 = T::now();
+		std::invoke(std::forward<decltype(f)>(f), std::forward<Args>(args)...);
+		auto t2 = T::now();
+		return std::chrono::duration<double, R>(t2 - t1);
+	}
+
 	LibcameraApp *app_;
 };
 
--- a/post_processing_stages/CMakeLists.txt
+++ b/post_processing_stages/CMakeLists.txt
@@ -25,6 +25,12 @@ else()
     message(WARNING "OpenCV not found, some stages will not be built!")
 endif()
 
+set(ENABLE_TFLITE)
+if (ENABLE_TFLITE)
+    set(SRC ${SRC} object_classify_tf_stage.cpp)
+    set(TARGET_LIBS ${TARGET_LIBS} tensorflow-lite atomic)
+endif()
+
 add_library(post_processing_stages ${SRC})
 target_link_libraries(post_processing_stages ${TARGET_LIBS})
 target_compile_definitions(post_processing_stages PUBLIC OPENCV_PRESENT=${OpenCV_FOUND})
--- a/post_processing_stages/object_classify_tf_stage.cpp
+++ b/post_processing_stages/object_classify_tf_stage.cpp
@@ -1,0 +1,370 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (C) 2021, Raspberry Pi (Trading) Limited
+ *
+ * negate_stage.cpp - image negate effect
+ */
+
+#include <algorithm>
+#include <fstream>
+#include <functional>
+#include <iostream>
+#include <map>
+#include <memory>
+#include <queue>
+#include <string>
+#include <vector>
+
+#include <libcamera/stream.h>
+
+#include "core/libcamera_app.hpp"
+#include "core/post_processing_stage.hpp"
+
+#include "tensorflow/lite/builtin_op_data.h"
+#include "tensorflow/lite/c/common.h"
+#include "tensorflow/lite/interpreter.h"
+#include "tensorflow/lite/kernels/register.h"
+#include "tensorflow/lite/string_util.h"
+
+using Stream = libcamera::Stream;
+
+struct TfSettings
+{
+	TfLiteType input_type = kTfLiteFloat32;
+	bool allow_fp16 = false;
+	tflite::string model_name = "/home/pi/models/mobilenet_quant_v1_224.tflite";
+	tflite::FlatBufferModel *model;
+	tflite::string labels_file_name = "/home/pi/models/labels.txt";
+	int number_of_threads = 4;
+	int number_of_results = 3;
+};
+
+constexpr int IMAGE_WIDTH = 224;
+constexpr int IMAGE_HEIGHT = 224;
+constexpr int IMAGE_CHANNELS = 3;
+
+class ObjectClassifyTfStage : public PostProcessingStage
+{
+public:
+	ObjectClassifyTfStage(LibcameraApp *app) : PostProcessingStage(app) {}
+
+	char const *Name() const override;
+
+	void Read(boost::property_tree::ptree const &params);
+
+	void Configure();
+
+	bool Process(CompletedRequest &completed_request);
+
+	void Stop();
+
+private:
+	void initialize();
+
+	void runInference();
+
+	TfLiteStatus readLabelsFile(const tflite::string &file_name, std::vector<tflite::string> *result,
+								size_t *found_label_count);
+
+	std::vector<uint8_t> yuvToRgb(uint8_t *mem, int w, int h, int stride);
+
+	void getTopResults(uint8_t *prediction, int prediction_size, size_t num_results,
+					   std::vector<std::pair<float, int>> *top_results, TfLiteType input_type);
+
+	Stream *stream_;
+	int w_, h_, stride_;
+	std::mutex future_ptr_mutex_;
+	std::mutex output_results_mutex_;
+	TfSettings settings_;
+	std::unique_ptr<tflite::FlatBufferModel> model_;
+	std::unique_ptr<tflite::Interpreter> interpreter_;
+	const std::vector<int> inputs_;
+	const std::vector<int> outputs_;
+	int input_;
+	std::vector<std::pair<std::string, float>> output_results_;
+	std::unique_ptr<std::future<void>> future_ptr_;
+	int refresh_rate_;
+	int top_n_results_;
+	float threshold_high_;
+	float threshold_low_;
+	std::string model_file_;
+	std::string labels_file_;
+	int display_labels_;
+	std::vector<tflite::string> labels_;
+	size_t label_count_;
+	std::vector<uint8_t> tensor_input_;
+	std::vector<std::pair<float, int>> top_results_;
+};
+
+#define NAME "object_classify_tf"
+
+char const *ObjectClassifyTfStage::Name() const
+{
+	return NAME;
+}
+
+void ObjectClassifyTfStage::Read(boost::property_tree::ptree const &params)
+{
+	top_n_results_ = params.get<int>("top_n_results", 3);
+	refresh_rate_ = params.get<int>("refresh_rate", 10);
+	threshold_high_ = params.get<float>("threshold_high", 0.01f);
+	threshold_low_ = params.get<float>("threshold_low", 0.01f);
+	model_file_ = params.get<std::string>("model_file", "/home/pi/models/mobilenet_quant_v1_224.tflite");
+	labels_file_ = params.get<std::string>("labels_file", "/home/pi/models/labels.txt");
+	display_labels_ = params.get<int>("display_labels", 1);
+
+	settings_.number_of_results = top_n_results_;
+	settings_.model_name = model_file_;
+	settings_.labels_file_name = labels_file_;
+
+	initialize();
+}
+
+void ObjectClassifyTfStage::initialize()
+{
+	model_ = tflite::FlatBufferModel::BuildFromFile(settings_.model_name.c_str());
+
+	settings_.model = model_.get();
+	std::cout << "Loaded model " << settings_.model_name << std::endl;
+	model_->error_reporter();
+
+	tflite::ops::builtin::BuiltinOpResolver resolver;
+
+	tflite::InterpreterBuilder(*model_, resolver)(&interpreter_);
+	if (!interpreter_)
+		throw std::runtime_error("Failed to construct interpreter");
+
+	std::cout << "Input size " << interpreter_->inputs().size() << std::endl;
+	interpreter_->SetAllowFp16PrecisionForFp32(settings_.allow_fp16);
+
+	if (settings_.number_of_threads != -1)
+		interpreter_->SetNumThreads(settings_.number_of_threads);
+
+	const std::vector<int> inputs_ = interpreter_->inputs();
+	const std::vector<int> outputs_ = interpreter_->outputs();
+
+	if (interpreter_->AllocateTensors() != kTfLiteOk)
+		throw std::runtime_error("Failed to allocate tensors");
+
+	if (readLabelsFile(settings_.labels_file_name, &labels_, &label_count_) != kTfLiteOk)
+		throw std::runtime_error("Failed to read labels file");
+}
+
+void ObjectClassifyTfStage::Configure()
+{
+	stream_ = app_->LoresStream();
+	if (stream_)
+	{
+		app_->StreamDimensions(stream_, &w_, &h_, &stride_);
+		if (w_ != IMAGE_WIDTH || h_ != IMAGE_HEIGHT)
+			throw std::runtime_error("Lores width and height must be 224");
+	}
+}
+
+bool ObjectClassifyTfStage::Process(CompletedRequest &completed_request)
+{
+	if (!stream_)
+		return false;
+
+	{
+		std::unique_lock<std::mutex> lck(future_ptr_mutex_);
+		if (completed_request.sequence % refresh_rate_ == 0 &&
+			(!future_ptr_ || future_ptr_->wait_for(std::chrono::seconds(0)) == std::future_status::ready))
+		{
+			libcamera::Span<uint8_t> buffer = app_->Mmap(completed_request.buffers[stream_])[0];
+			uint8_t *ptr = buffer.data();
+
+			tensor_input_ = yuvToRgb(ptr, w_, h_, stride_);
+
+			future_ptr_ = std::make_unique<std::future<void>>();
+			*future_ptr_ = std::move(std::async(std::launch::async, [this] {
+				auto time_taken =
+					ExecutionTime<std::micro>(std::bind(&ObjectClassifyTfStage::runInference, this)).count();
+				std::cout << time_taken << "ms" << std::endl;
+			}));
+		}
+	}
+
+	std::unique_lock<std::mutex> lock(output_results_mutex_);
+	std::string annotation = "Detected";
+
+	for (const auto &result : output_results_)
+	{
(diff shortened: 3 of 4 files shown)


RaspiShonak:
@naushir @davidplowman

core/post_processing_stage.hpp
```diff
@@ -45,6 +46,19 @@ class PostProcessingStage
 	virtual void Teardown();
 
 protected:
+	// Helper to calculate the execution time of any callable object and return it in as a std::chrono::duration.
+	// For functions returning a value, the simplest thing would be to wrap the call in a lambda and capture
+	// the return value.
+	template <class R = std::micro, class T = std::chrono::steady_clock, class F, class... Args>
+
+	static auto ExecutionTime(F &&f, Args &&... args)
```
naushir, line 54: Extra line can be removed.

post_processing_stages/CMakeLists.txt
```diff
@@ -25,6 +25,12 @@ else()
     message(WARNING "OpenCV not found, some stages will not be built!")
 endif()
 
+set(ENABLE_TFLITE)
+if (ENABLE_TFLITE)
+    set(SRC ${SRC} object_classify_tf_stage.cpp)
+    set(TARGET_LIBS ${TARGET_LIBS} tensorflow-lite atomic)
+endif()
+
```
naushir, line 33: Do we want to have a mechanism where we check if the tensorflow-lite library is found, similar to what we do for OpenCV? 

Is atomic needed?

post_processing_stages/object_classify_tf_stage.cpp
```diff
@@ -0,0 +1,370 @@
...
+			libcamera::Span<uint8_t> buffer = app_->Mmap(completed_request.buffers[stream_])[0];
+			uint8_t *ptr = buffer.data();
+
+			tensor_input_ = yuvToRgb(ptr, w_, h_, stride_);
+
+			future_ptr_ = std::make_unique<std::future<void>>();
+			*future_ptr_ = std::move(std::async(std::launch::async, [this] {
+				auto time_taken =
+					ExecutionTime<std::micro>(std::bind(&ObjectClassifyTfStage::runInference, this)).count();
+				std::cout << time_taken << "ms" << std::endl;
```
davidplowman, line 183: I think we should probably read a parameter that determines whether to print the timings - not everyone will want it.

post_processing_stages/object_classify_tf_stage.cpp
```diff
@@ -0,0 +1,370 @@
...
+
+	settings_.model = model_.get();
+	std::cout << "Loaded model " << settings_.model_name << std::endl;
+	model_->error_reporter();
+
+	tflite::ops::builtin::BuiltinOpResolver resolver;
+
+	tflite::InterpreterBuilder(*model_, resolver)(&interpreter_);
+	if (!interpreter_)
+		throw std::runtime_error("Failed to construct interpreter");
```
davidplowman, line 135: I think we should normally try and make the error messages say more clearly where they come from. So this one would look like
`throw std::runtime_error("ObjectClassifyTfStage: failed to construct interpreter");`
(And the same for all the other ones...)

post_processing_stages/object_classify_tf_stage.cpp
```diff
@@ -0,0 +1,370 @@
...
+
+	std::cout << "Input size " << interpreter_->inputs().size() << std::endl;
+	interpreter_->SetAllowFp16PrecisionForFp32(settings_.allow_fp16);
```
naushir, line 138: Perhaps we have a "log" option in the settings and only do the std::cout if that is set?
I don't think we want to write to stdout unconditionally.

post_processing_stages/object_classify_tf_stage.cpp
```diff
@@ -0,0 +1,370 @@
...
+	{
+		std::unique_lock<std::mutex> lck(future_ptr_mutex_);
+		if (completed_request.sequence % refresh_rate_ == 0 &&
+			(!future_ptr_ || future_ptr_->wait_for(std::chrono::seconds(0)) == std::future_status::ready))
+		{
+			libcamera::Span<uint8_t> buffer = app_->Mmap(completed_request.buffers[stream_])[0];
+			uint8_t *ptr = buffer.data();
+
+			tensor_input_ = yuvToRgb(ptr, w_, h_, stride_);
+
```
naushir, line 178: Would this be more efficient if placed in the std::async? Not really sure it matters.

post_processing_stages/object_classify_tf_stage.cpp
```diff
@@ -0,0 +1,370 @@
...
+			future_ptr_ = std::make_unique<std::future<void>>();
+			*future_ptr_ = std::move(std::async(std::launch::async, [this] {
+				auto time_taken =
+					ExecutionTime<std::micro>(std::bind(&ObjectClassifyTfStage::runInference, this)).count();
+				std::cout << time_taken << "ms" << std::endl;
+			}));
```
naushir, line 184: Again, we probably don't want to write this unconditionally.

post_processing_stages/object_classify_tf_stage.cpp
```diff
@@ -0,0 +1,370 @@
...
+		}
+	}
+
+	std::unique_lock<std::mutex> lock(output_results_mutex_);
+	std::string annotation = "Detected";
+
+	for (const auto &result : output_results_)
+	{
+		annotation = annotation + ", " + result.first + " : " + std::to_string(result.second);
+	}
```
davidplowman, line 194: I think we could improve this a bit. If there's nothing detected we say "Detected: (nothing)", if there are several things "Detected: cat 73%, dog 61% or something like that without any extraneous commas or anything.

post_processing_stages/object_classify_tf_stage.json
```diff
@@ -0,0 +1,22 @@
...
+    "object_classify_tf":
+    {
+    "top_n_results" : 2,
+    "refresh_rate" : 30,
+    "threshold_high" : 0.6,
+    "threshold_low" : 0.4,
+    "model_file" : "/home/pi/models/mobilenet_v1_1.0_224_quant.tflite",
+    "labels_file" : "/home/pi/models/labels.txt",
+    "display_labels" : 1
+    },
```
naushir, line 11: Indentation?

post_processing_stages/object_classify_tf_stage.cpp
```diff
@@ -0,0 +1,370 @@
...
+	uint8_t *Y = mem, *U = mem + h * stride, *V = mem + (5 / 4) * h * stride;
+	int dst;
+	std::vector<uint8_t> output(h * w * 3);
+
+	for (int y = 0; y < h; y++)
+	{
+		for (int x = 0; x < w; x++)
+		{
+			Y = mem + y * stride + x;
+			U = mem + (h * stride) + ((int)floor(y / 2)) * (stride / 2) + ((int)floor(x / 2));
```
davidplowman, line 280: I suspect all the "(int)floor(..)" stuff is unnecessary as integer arithmetic should just gives us what we want.

post_processing_stages/object_classify_tf_stage.cpp
```diff
@@ -0,0 +1,370 @@
...
+			V = mem + (5 * h * stride) / 4 + ((int)floor(y / 2)) * (stride / 2) + ((int)floor(x / 2));
```
davidplowman, line 281: Is this just `V = U + h * stride / 4;` maybe?

post_processing_stages/object_classify_tf_stage.cpp
```diff
@@ -0,0 +1,370 @@
...
+		result->emplace_back();
+	}
+
+	return kTfLiteOk;
+}
+
+std::vector<uint8_t> ObjectClassifyTfStage::yuvToRgb(uint8_t *mem, int w, int h, int stride)
+{
+	uint8_t *Y = mem, *U = mem + h * stride, *V = mem + (5 / 4) * h * stride;
+	int dst;
```
naushir, line 272: (5 / 4) == 1 with integer truncation.

post_processing_stages/object_classify_tf_stage.cpp
```diff
@@ -0,0 +1,370 @@
...
+	std::vector<uint8_t> output(h * w * 3);
+
+	for (int y = 0; y < h; y++)
+	{
+		for (int x = 0; x < w; x++)
+		{
+			Y = mem + y * stride + x;
+			U = mem + (h * stride) + ((int)floor(y / 2)) * (stride / 2) + ((int)floor(x / 2));
+			V = mem + (5 * h * stride) / 4 + ((int)floor(y / 2)) * (stride / 2) + ((int)floor(x / 2));
+			dst = (x + y * w) * 3;
```
naushir, line 282: Integer divisions here probably not correct.

post_processing_stages/object_classify_tf_stage.cpp
```diff
@@ -0,0 +1,370 @@
...
+
+void ObjectClassifyTfStage::getTopResults(uint8_t *prediction, int prediction_size, size_t num_results,
+										  float threshold_high_, float threshold_low_,
+										  std::vector<std::pair<float, int>> *top_results, TfLiteType input_type)
+{
+	// Will contain top N results in ascending order.
+	std::priority_queue<std::pair<float, int>, std::vector<std::pair<float, int>>, std::greater<std::pair<float, int>>>
+		top_result_pq;
+	std::vector<std::pair<float, int>> top_results_old = top_results_;
+	top_results_.clear();
```
davidplowman, line 303: Does this want a `std::move(top_results_)`?

post_processing_stages/object_classify_tf_stage.cpp
```diff
@@ -0,0 +1,370 @@
...
+
+	const long count = prediction_size; // NOLINT(runtime/int)
```
davidplowman, line 305: I'm guessing we don't need the "NOLINT" stuff here...

post_processing_stages/object_classify_tf_stage.cpp
```diff
@@ -0,0 +1,370 @@
...
+	}
+
+	// Copy to output vector and reverse into descending order.
+	while (!top_result_pq.empty())
+	{
+		top_results->push_back(top_result_pq.top());
+		top_result_pq.pop();
+	}
+	std::reverse(top_results->begin(), top_results->end());
+}
```
naushir, line 364: This seems a bit complicated for what (i think) it is doing.  Perhaps just filling a vector with all results (provided the confidence is above the threshold), and calling std::sort() at the end might be better?  The caller then only needs to look at the top N results.

post_processing_stages/object_classify_tf_stage.cpp
```diff
@@ -0,0 +1,370 @@
...
+	}
+
+	std::unique_lock<std::mutex> lock(output_results_mutex_);
+	std::string annotation = "Detected";
+
+	for (const auto &result : output_results_)
+	{
+		annotation = annotation + ", " + result.first + " : " + std::to_string(result.second);
+	}
+
```
naushir, line 195: It might be more efficient to use a std::stringstream to construct the annotation string.

naushir (commented):
Looks good, mostly cosmetic suggestions, but there are one or two things to look at in more detail.

post_processing_stages/CMakeLists.txt
davidplowman: I agree that we should check if it's there, though the default is not to include TF stages so it's more of a nice-to-have.

libatomic did seem to be necessary, but it's worth checking again seeing as various bits of code have been chucked out!

post_processing_stages/object_classify_tf_stage.cpp
davidplowman: I don't mind too much here as it's a one-off, but I agree with your later similar one.

post_processing_stages/object_classify_tf_stage.cpp
davidplowman: Good spot! Though looking further down it seems these values don't get used. Probably best just to delete the initialisations here.

post_processing_stages/object_classify_tf_stage.cpp
davidplowman: I think that sounds good. Then there's a final step of adding back any of the old results that are not already in the new list, but above the lower threshold. (It's a Schmitt trigger kind of thing.)

post_processing_stages/object_classify_tf_stage.cpp
```diff
@@ -0,0 +1,370 @@
...
+	Stream *stream_;
+	int w_, h_, stride_;
+	std::mutex future_ptr_mutex_;
+	std::mutex output_results_mutex_;
+	TfSettings settings_;
+	std::unique_ptr<tflite::FlatBufferModel> model_;
+	std::unique_ptr<tflite::Interpreter> interpreter_;
+	const std::vector<int> inputs_;
+	const std::vector<int> outputs_;
+	int input_;
```
davidplowman, line 83: I suspect that this `input_` doesn't get used, is that right? If so we should delete it.

post_processing_stages/object_classify_tf_stage.cpp
```diff
@@ -0,0 +1,370 @@
...
+static PostProcessingStage *Create(LibcameraApp *app)
+{
+	return new ObjectClassifyTfStage(app);
+}
+
+static RegisterStage reg(NAME, &Create);
+
+void ObjectClassifyTfStage::runInference()
+{
+	int input_ = interpreter_->inputs()[0];
```
davidplowman, line 213: This is a local, not a member, variable. As such it should be just `input`.

post_processing_stages/object_classify_tf_stage.cpp
```diff
@@ -0,0 +1,370 @@
...
+
+	Stream *stream_;
+	int w_, h_, stride_;
+	std::mutex future_ptr_mutex_;
+	std::mutex output_results_mutex_;
+	TfSettings settings_;
+	std::unique_ptr<tflite::FlatBufferModel> model_;
+	std::unique_ptr<tflite::Interpreter> interpreter_;
+	const std::vector<int> inputs_;
+	const std::vector<int> outputs_;
```
davidplowman, line 82: I don't think we're using either inputs_ or outputs_? They might want to be deleted.

post_processing_stages/object_classify_tf_stage.cpp
```diff
@@ -0,0 +1,370 @@
...
+		throw std::runtime_error("Failed to construct interpreter");
+
+	std::cout << "Input size " << interpreter_->inputs().size() << std::endl;
+	interpreter_->SetAllowFp16PrecisionForFp32(settings_.allow_fp16);
+
+	if (settings_.number_of_threads != -1)
+		interpreter_->SetNumThreads(settings_.number_of_threads);
+
+	const std::vector<int> inputs_ = interpreter_->inputs();
+	const std::vector<int> outputs_ = interpreter_->outputs();
```
davidplowman, line 144: I'm not sure these lines are doing anything, but it would be worth checking. I expect they just make a local copy of some stuff which then just gets discarded. (They're also probably hiding the class member of the same name, though I'm suspicious that those can be deleted too.)

post_processing_stages/object_classify_tf_stage.cpp
```diff
@@ -0,0 +1,370 @@
...
+{
+	return new ObjectClassifyTfStage(app);
+}
+
+static RegisterStage reg(NAME, &Create);
+
+void ObjectClassifyTfStage::runInference()
+{
+	int input_ = interpreter_->inputs()[0];
+	settings_.input_type = interpreter_->tensor(input_)->type;
```
davidplowman, line 214: I'm wondering a bit whether input_type needs to be a field in the settings_ object. It looks to me like we set it unconditionaly here and then just pass it to getTopResults, at which point it could certainly be at most a local variable?

post_processing_stages/object_classify_tf_stage.cpp
```diff
@@ -0,0 +1,370 @@
...
+}
+
+bool ObjectClassifyTfStage::Process(CompletedRequest &completed_request)
+{
+	if (!stream_)
+		return false;
+
+	{
+		std::unique_lock<std::mutex> lck(future_ptr_mutex_);
+		if (completed_request.sequence % refresh_rate_ == 0 &&
```
davidplowman, line 171: Might be worth checking somewhere that refresh_rate isn't zero.

post_processing_stages/object_classify_tf_stage.cpp
RaspiShonak: Verbose option added to toggle execution time and detection results from being printed.

post_processing_stages/object_classify_tf_stage.cpp
RaspiShonak: Added Stage name for all.

post_processing_stages/object_classify_tf_stage.cpp
RaspiShonak: removed int floors.

post_processing_stages/object_classify_tf_stage.cpp
RaspiShonak: Yep, that works! changed it to the equation above.

post_processing_stages/object_classify_tf_stage.cpp
RaspiShonak: changed it to std::move

post_processing_stages/object_classify_tf_stage.cpp
RaspiShonak: Removed the NOLINT

post_processing_stages/CMakeLists.txt
RaspiShonak: Atomic actually not needed anymore!

post_processing_stages/object_classify_tf_stage.cpp
RaspiShonak: changed to be conditional on verbose

post_processing_stages/object_classify_tf_stage.cpp
RaspiShonak: Deleted that bit.

post_processing_stages/object_classify_tf_stage.cpp
RaspiShonak: Deleted both of them.

post_processing_stages/object_classify_tf_stage.cpp
RaspiShonak: Actually not needed at all. Both have been deleted.

post_processing_stages/object_classify_tf_stage.cpp
RaspiShonak: Added if (refresh_rate_ && ... ){

post_processing_stages/object_classify_tf_stage.cpp
RaspiShonak: Yep, not needed at all. Both have been deleted.

post_processing_stages/object_classify_tf_stage.cpp
RaspiShonak: Awesome, changed it to std::stringstream

post_processing_stages/CMakeLists.txt
davidplowman: We still seem to have `atomic` on line 31... was it supposed to be removed?

Issue #66, opened 2021-08-13
The preview image is wrong when the preview width is greater than 2048 using Pi3.


glddiv:
When using libcamera-still to set the width of the preview resolution (--viewfinder-width) greater than 2048 on Pi3, the image is not displayed correctly. (The saved picture is normal)
The effect is as follows:
![crop](https://user-images.githubusercontent.com/24386338/129297536-a23ee40b-b082-465d-899e-82cb326f8fa8.png)

There is no problem using the same SD Card on Pi4.

kernel version:
`Linux raspberrypi 5.10.52-v7+ #1444 SMP Thu Aug 12 20:22:57 BST 2021 armv7l GNU/Linux`

firmware version:
```
Aug 11 2021 14:01:51 
Copyright (c) 2012 Broadcom
version 5c8758dbac5b27911d2384c5dcd5db843528b606 (clean) (release) (start_x)
```

naushir:
Could you provide the exact command line you are using to get this corruption?  Additionally, what sensor is this on?
Providing any log output from libcamera would also be helpfuly.

glddiv:
Sorry for that. I thought this might be a known issue or limitation.
libcamera-still:
```
pi@raspberrypi:~/libcamera-apps/build $ ./libcamera-still -t 0 --viewfinder-width=2592 --viewfinder-height=1944
[0:02:52.656226158] [1281]  INFO Camera camera_manager.cpp:294 libcamera v0.0.0+2880-9947b4cf
[0:02:53.057193702] [1281]  INFO Camera camera.cpp:870 configuring streams: (0) 2592x1944-YUV420
[0:02:53.058737348] [1283]  INFO RPI raspberrypi.cpp:121 Mode: 3280x2464 fmt RG10 Score: 1808.49 (best 1808.49)
[0:02:53.059017396] [1283]  INFO RPI raspberrypi.cpp:121 Mode: 1920x1080 fmt RG10 Score: 4738.67 (best 1808.49)
[0:02:53.059145647] [1283]  INFO RPI raspberrypi.cpp:121 Mode: 1640x1232 fmt RG10 Score: 4834.49 (best 1808.49)
[0:02:53.059266501] [1283]  INFO RPI raspberrypi.cpp:121 Mode: 640x480 fmt RG10 Score: 8332 (best 1808.49)
[...]

PR #67, opened 2021-08-18
Limit viewfinder image size by the preview window capabilities

davidplowman:
Some devices can only render limited size images into the preview
window (for Pi 3, this limit is 2048x2048). We now fetch these limits,
and ensure that the viewfinder mode doesn't request anything larger
from the camera/ISP.

### DIFF
--- a/core/libcamera_app.cpp
+++ b/core/libcamera_app.cpp
@@ -144,6 +144,16 @@ void LibcameraApp::ConfigureViewfinder()
 			std::cout << "Viewfinder size chosen is " << size.toString() << std::endl;
 	}
 
+	// Finally trim the image size to the largest that the preview can handle.
+	Size max_size;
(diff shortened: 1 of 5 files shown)


davidplowman:
@naushir This mostly fixes https://github.com/raspberrypi/libcamera-apps/issues/66 as you'll no longer get a garbled preview window. Though there are still some sticky corner cases (e.g. recording a video on a Pi 3 - it woudl have to be MJPEG - wider than 2048 and wanting to see a preview).

davidplowman:
I discussed this one with folks yesterday. There's still this slight strangeness where GL claims a maximum texture size of 4096 and DRM a max display width of 7680, but it still seems like this is a reasonable thing to do. In particular it solves the obvious garbled-preview-on-Pi-3 problem.

PR #68, opened 2021-08-18
post-processing: Add a motion_detect stage

davidplowman:
The motion_detect stage looks for pixel differences in the low
resolution image, which should normally be configured quite small,
such as 128x96 pixels. If enough pixels have changed sufficiently,
this counts as "motion".

### DIFF
--- a/post_processing_stages/motion_detect_stage.cpp
+++ b/post_processing_stages/motion_detect_stage.cpp
@@ -1,0 +1,190 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (C) 2021, Raspberry Pi (Trading) Limited
+ *
+ * motion_detect_stage.cpp - motion detector
+ */
+
+// A simple motion detector. It needs to be given a low resolution image and it
+// compares pixels in the current low res image against the value in the corresponding
+// location in the previous one. If it exceeds a threshold it gets counted as
+// "different". If enough pixels are different, that indicates "motion".
+// A low res image of something like 128x96 is probably more than enough, and you
+// can always subsample with hskip and vksip.
+
+// Because this gets run in parallel by the post-processing framework, it means
+// the "previous frame" is not totally guaranteed to be the actual previous one,
+// though in practice it is, and it doesn't actually matter even if it wasn't.
+
+// The stage adds "motion_detect.result" to the metadata. When this claims motion,
+// the application can take that as true immediately. To be sure there's no motion,
+// an application should probably wait for "a few frames" of "no motion".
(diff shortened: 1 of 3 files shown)


davidplowman:
@naushir

post_processing_stages/motion_detect_stage.cpp
```diff
@@ -0,0 +1,190 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (C) 2021, Raspberry Pi (Trading) Limited
+ *
+ * motion_detect_stage.cpp - motion detector
+ */
+
+// A simple motion detector. It needs to be give a low resolution image and it
+// compares pixels in the current low res image again the value in the corresponding
```
naushir, line 9: s/again/against/

post_processing_stages/motion_detect_stage.cpp
```diff
@@ -0,0 +1,190 @@
...
+// location in the previous one. If it exceeds a threshold it gets counted as
+// "different". If enough pixels are different, that indicates "motion".
+// A low res image of something like 128x96 is probably more than enough, and you
+// can always subsample with hskip and vksip.
+
+// Because this gets run in parallel by the post-processing framework, it means
+// the "previous frame" is not totally guaranteed to be the actual previous one,
+// though in practice it is, and it doesn't actually matter even if it wasn't.
+
```
naushir, line 18: Not something to do now, but I wonder if we should have a mode that ensures we use the "current" and "previous" frame only.  Would that be useful?

naushir (commented):
Apart from the small typo, it looks good!

post_processing_stages/motion_detect_stage.cpp
davidplowman: I agree it would be nice to guarantee that the "previous" frame really is the actual genuinely previous one. But we're at the mercy of the multiple threads, which thread tries to run first, who gets the mutex first... that kind of thing. So I'm not entirely sure what the best way is to do that. Perhaps run it before we split into several threads? Does that over-complicate the post-processing idea? Open to suggestions on this one!

naushir: Need to think about how to do this a bit more.  Given the postprocessing thread dispatch code ensures that we return results in order, I think we may be able to do some trickery with mutexes or other sync primitives...

post_processing_stages/motion_detect_stage.cpp
davidplowman: And s/give/given/ too!

PR #69, opened 2021-08-23
Qt preview

davidplowman:


davidplowman:
@naushir This provides a preview window implemented with Qt. Why on earth would we want that, especially when it involves buffer copying, CPU resizing and colour conversions...? Well, it works with X forwarding, so it does provide an option for people who really feel they have to have a local preview window when the Pi running the camera is somewhere else.

preview/preview.cpp
```diff
@@ -0,0 +1,52 @@
...
+Preview *make_egl_preview(Options const *options);
+Preview *make_drm_preview(Options const *options);
+Preview *make_qt_preview(Options const *options);
+
+Preview *make_preview(Options const *options)
+{
+	if (options->nopreview)
+		return make_null_preview(options);
+	else if (options->qt_preview)
+		return make_qt_preview(options);
```
naushir, line 20: Does this need an ``#if QT_PRESET`` type guard?

preview/preview.cpp
```diff
@@ -0,0 +1,52 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (C) 2020, Raspberry Pi (Trading) Ltd.
+ *
```
naushir, line 4: 2021 :)

preview/drm_preview.cpp
```diff
@@ -20,7 +20,7 @@ class DrmPreview : public Preview
 	~DrmPreview();
 	// Display the buffer. You get given the fd back in the BufferDoneCallback
 	// once its available for re-use.
-	virtual void Show(int fd, size_t size, int width, int height, int stride) override;
+	virtual void Show(int fd, void *mem, size_t size, int width, int height, int stride) override;
 	// Reset the preview window, clearing the current buffers and being ready to
```
naushir, line 24: Not too fussed about this, but perhaps we should use a Span for consistency?

preview/qt_preview.cpp
```diff
@@ -0,0 +1,175 @@
...
+		uint8_t *Y_start = (uint8_t *)mem;
+		uint8_t *U_start = Y_start + stride * height;
+		int uv_size = (stride / 2) * (height / 2);
+		uint8_t *dest = pane_->image.bits();
+
+		// Possibly this should be locked in case a repaint is happening? In practice the risk
+		// is only that there might be some tearing, so I don't think we worry. We could speed
+		// it up by getting the ISP to supply RGB, but I'm not sure I want to handle that extra
+		// possibility in our main application code, so we'll put up with the slow conversion.
+		for (int y = 0; y < window_height_; y++)
```
naushir, line 93: Perhaps also an assert to make sure we are coming into here with a YUV420 format image?

preview/preview.cpp
davidplowman: Yes!

preview/drm_preview.cpp
davidplowman: Yes, I think that makes sense. Our post-processing branch has switched over to using Spans more so this will fit well (though probably cause a minor merge conflict!).

preview/qt_preview.cpp
davidplowman: All the preview windows only support YUV420 so we probably want to check back in the LibcameraApp::previewThread, but doing so would clearly be sensible.

davidplowman:
@naushir I think this latest version takes care of those various things you pointed out - let me know what you think!

naushir:
Looks good!

PR #71, opened 2021-09-01
Tf stage

davidplowman:
@naushir 
This set of patches adds the new TfStage base class which makes the addition of new TFLite-based post-processing stages easier. The first of the commits adds the new TfStage class implementation.

Subsequently there's one patch that refactors the Object Classifier to use the new base class, and another that further refactors the Pose Estimator in the same way. (There's a final small file renaming patch that just fixes up an incorrect JSON file name.,)


post_processing_stages/tf_stage.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+	readExtras(params);
+}
+
+void TfStage::initialise()
+{
+	model_ = tflite::FlatBufferModel::BuildFromFile(config_->model_file.c_str());
+	std::cout << "TfStage: Loaded model " << config_->model_file << std::endl;
+	model_->error_reporter();
+
```
naushir, line 33: If there is an error (e.g. model file missing), do we need some kind of exception here?

naushir (commented):
Looks good, only one minor thing about the models file.

post_processing_stages/tf_stage.cpp
davidplowman: Yes, I'll fix that.

PR #74, opened 2021-09-06
libcamera-still: Add option for immediate capture

davidplowman:
The "immediate" option causes libcamera-still to capture the very
first frame from the camera, with no initial preview phase at all.


davidplowman:
@naushir One more pull request... it really was surprisingly trivial!

PR #76, opened 2021-09-08
Cope with libcamera now returning single plane buffers as if multi plane

davidplowman:
Two small tweaks are required:

- in libcamera_app.cpp we must spot planes with the same fd and
  accumulate them so as to mmap the buffer as a single plane

- in libcamera_encoder.hpp we must be careful to pass the full buffer
  size to the encoder not the plane's length (which is now just the Y
  portion)


davidplowman:
@naushir With these fixes everything seems to have recovered using top-of-tree libcamera!

PR #78, opened 2021-09-09
build: fix compiler error and warnings with GCC 10

davidplowman:
Two small fix-ups to prevent one failure and some Boost warnings when
building with GCC 10.3.0. Harmless with earlier versions.

### DIFF
--- a/core/post_processing_stage.hpp
+++ b/core/post_processing_stage.hpp
@@ -5,7 +5,11 @@
  * post_processing_stage.hpp - Post processing stage base class definition.
  */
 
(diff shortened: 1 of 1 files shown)


davidplowman:
@naushir I've tried this and it fixes both the hard error and the Boost warnings that I see in GCC 10.3.0 (but not in 8.3.0). I would ping Serge but can't find his Github id...

naushir:
LGTM

@XECDesign

XECDesign:
That fixed it, thanks.

PR #79, opened 2021-09-14
apps: Add a new libcamera-detect app

davidplowman:
This uses the post-processing stage ObjectDetectTfStage to identify
named objects and capture JPEGs when it spots them.

Example: libcamera-detect --post-process-file object_detect_tf.json --lores-width 400 --lores-height 300 -t 0 --object cat -o cat%03d.jpg


davidplowman:
@naushir More fun with TensorFlow... this app runs preview until the named object is detected, then does a full resolution JPEG capture, before returning to preview.

PR #82, opened 2021-09-16
test: Make test.py separate the output directory from the executable …

davidplowman:
…locations

It defaults to the same behaviour as before but now it's much more
convenient for testing (for example) executables that have been
installed, e.g.

cd
mkdir outputs
libcamera-apps/test.py --exe-dir /usr/bin --output-dir outputs

(Also fixed some historic code formatting errors that had been ignored
previously.)


davidplowman:
@naushir No urgency on this, just something that was necessary for testing the installed executables in our candidate Bullseye images.

PR #85, opened 2021-09-21
Install libraries as per GNU Coding Standards

dwrobel:
Signed-off-by: Damian Wrobel <dwrobel@ertelnet.rybnik.pl>

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -1,9 +1,11 @@
 cmake_minimum_required(VERSION 3.6)
 
+include(GNUInstallDirs)
+
 find_package(Boost REQUIRED COMPONENTS program_options)
 
 add_library(libcamera_app libcamera_app.cpp post_processing_stage.cpp post_processor.cpp)
 set_target_properties(libcamera_app PROPERTIES PREFIX "" IMPORT_PREFIX "")
 target_link_libraries(libcamera_app pthread images preview ${LIBCAMERA_LINK_LIBRARIES} ${Boost_LIBRARIES} post_processing_stages)
 
-install(TARGETS libcamera_app LIBRARY DESTINATION lib ARCHIVE DESTINATION lib)
+install(TARGETS libcamera_app LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR} ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR})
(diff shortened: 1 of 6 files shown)


naushir:
This seems like a reasonable change to make.
Just wondering, should we also use `CMAKE_INSTALL_BINDIR` for the executable to keep things consistent?

dwrobel:
> This seems like a reasonable change to make.
> Just wondering, should we also use `CMAKE_INSTALL_BINDIR` for the executable to keep things consistent?

I'm not observing any problems with binaries.

Here is the build attempt without this patch: https://copr.fedorainfracloud.org/coprs/dwrobel/libcamera-apps/build/2830959/
Here is the build with this patch applied: https://copr.fedorainfracloud.org/coprs/dwrobel/libcamera-apps/build/2831036/

As you might observe the problem only exists on aarch64 and x86_64 platforms and is applicable to the following files:
```
RPM build errors:
    File not found: /builddir/build/BUILDROOT/libcamera-apps-0.0.0-0.1.20210915git4868f69.fc33.aarch64/usr/lib64/libcamera_app.so
    File not found: /builddir/build/BUILDROOT/libcamera-apps-0.0.0-0.1.20210915git4868f69.fc33.aarch64/usr/lib64/libencoders.so
    File not found: /builddir/build/BUILDROOT/libcamera-apps-0.0.0-0.1.20210915git4868f69.fc33.aarch64/usr/lib64/libimages.so
    File not found: /builddir/build/BUILDROOT/libcamera-apps-0.0.0-0.1.20210915git4868f69.fc33.aarch64/usr/lib64/liboutputs.so
    File not found: /builddir/build/BUILDROOT/libcamera-apps-0.0.0-0.1.20210915git4868f69.fc33.aarch64/usr/lib64/libpost_processing_stages.so
    File not found: /builddir/build/BUILDROOT/libcamera-apps-0.0.0-0.1.20210915git4868f69.fc33.aarch64/usr/lib64/libpreview.so
```
that's why this patch is limited to libraries only.

naushir:
@XECDesign is this something that will trip you up?

davidplowman:
I've tried this branch out on one of our aarch64 Bullseye pre-release images - seems to work fine. (To be honest I couldn't see any difference in where the libraries went... just wondering if that is expected?)

XECDesign:
> @XECDesign is this something that will trip you up?

Nope, shouldn't.

dwrobel:
>just wondering if that is expected?

You might observe a difference only if you or your Linux distribution, during packaging into .rpm, deb, passes value for `${CMAKE_INSTALL_LIBDIR}`. See: https://cmake.org/cmake/help/latest/command/install.html.

PR #87, opened 2021-09-24
libcamera_app: Consolidate move operations

naushir:
Mark CompletedRequest as non-movable.
Fixup Metadata::Set() forwarding the metadata value correctly.


naushir:
@davidplowman

PR #88, opened 2021-09-27
libcamera-detect: Fix bug shown up by CompleteRequest move vs. copy f…

davidplowman:
…ixes

The CompletedRequest gets (correctly) moved now rather than copied, so
you can't look in the metadata after sending it to the preview, you
must do it before.


davidplowman:
@naushir This is the fix for my cat spy-cam!

PR #89, opened 2021-09-27
Version information for libcamera-apps

naushir:
@davidplowman, might be worth giving this a go on your environment to make sure it works correctly!


davidplowman:
Works well for me!

Issue #90, opened 2021-09-28
libcamera-jpeg SIGABRT @libcamera::CameraConfiguration::at


kbingham:
From: Damian Wrobel 2021-09-21 19:01:03 CEST : https://bugs.libcamera.org/show_bug.cgi?id=78

I'm observing a SIGABRT @libcamera::CameraConfiguration::at when using libcamera-jpeg application as follows:

$ libcamera-jpeg -o test.jpg
Preview window unavailable
[144:40:33.264346482] [28470]  INFO Camera camera_manager.cpp:294 libcamera v0.0.0+1-dab03790
[144:40:33.335326696] [28470]  INFO Camera camera.cpp:937 configuring streams: (0) 1640x1232-YUV420
[144:40:33.335481490] [28474]  INFO RPI raspberrypi.cpp:122 Mode: 3280x2464 fmt RG10 Score: 2218 (best 2218)
[144:40:33.335544174] [28474]  INFO RPI raspberrypi.cpp:122 Mode: 1920x1080 fmt RG10 Score: 2041.48 (best 2041.48)
[144:40:33.335579655] [28474]  INFO RPI raspberrypi.cpp:122 Mode: 1640x1232 fmt RG10 Score: 1500 (best 1500)
[144:40:33.335612118] [28474]  INFO RPI raspberrypi.cpp:122 Mode: 640x480 fmt RG10 Score: 5004.81 (best 1500)
[144:40:33.335666376] [28474]  INFO RPI raspberrypi.cpp:122 Mode: 3280x2464 fmt pRAA Score: 1718 (best 1500)
[144:40:33.335696728] [28474]  INFO RPI raspberrypi.cpp:122 Mode: 1920x1080 fmt pRAA Score: 1541.48 (best 1500)
[144:40:33.335728542] [28474]  INFO RPI raspberrypi.cpp:122 Mode: 1640x1232 fmt pRAA Score: 1000 (best 1000)
[144:40:33.335768856] [28474]  INFO RPI raspberrypi.cpp:122 Mode: 640x480 fmt pRAA Score: 4504.81 (best 1000)
[144:40:33.335805430] [28474]  INFO RPI raspberrypi.cpp:122 Mode: 3280x2464 fmt RGGB Score: 3218 (best 1000)
[144:40:33.335834615] [28474]  INFO RPI raspberrypi.cpp:122 Mode: 1920x1080 fmt RGGB Score: 3041.48 (best 1000)
[...]

PR #91, opened 2021-09-28
core: Fix unsafe memory access in ConfigureStill

davidplowman:
It wasn't checking that the raw stream existed before accessing it.


davidplowman:
@naushir @kbingham I think this is probably the fix for https://github.com/raspberrypi/libcamera-apps/issues/90

PR #99, opened 2021-09-29
Update to version script.

naushir:


naushir:
@davidplowman small change to stop using utf-8 and allow the system locale.

davidplowman:
And some improved error catching too, I see!

PR #100, opened 2021-09-30
Use shared_ptr to pass CompletedRequests around

davidplowman:
Unfortunately this is one big all-or-nothing change. There's a lot of
mundane CompletedRequest -> CompletedRequestPtr changes here, but some
are more significant.

* The shared_ptr has a custom deleter so that once all references to
  it are dropped it calls LibcameraApp::queueRequest for
  recycling. (There's a nasty corner case about a CompletedRequest
  being recycled once the camera has been stopped and restarted.)

* LibcameraApp::ShowPreview needs to take a reference to the shared_ptr.

* LibcameraEncoder::EncodeBuffer needs to take a reference to the
  shared_ptr.

* The post-processor can re-use the shared_ptr both when it receives
  the CompletedRequest, and when it gets passed on to the message
  queue for the application.

* Neither the encoder nor the preview need a callback anymore to pass
  the buffer on (and ultimately ensure that LibcameraApp::QueueRequest
  is called, as this now happens automagically).

In general, I've tried to re-use references (by moving the shared_ptr)
wherever it's appropriate.

### DIFF
--- a/apps/libcamera_detect.cpp
+++ b/apps/libcamera_detect.cpp
@@ -55,7 +55,6 @@ static void event_loop(LibcameraDetectApp &app)
 	app.OpenCamera();
 	app.ConfigureViewfinder();
 	app.StartCamera();
-	app.SetPreviewDoneCallback(std::bind(&LibcameraApp::QueueRequest, &app, std::placeholders::_1));
 	auto start_time = std::chrono::high_resolution_clock::now();
 	unsigned int last_capture_frame = 0;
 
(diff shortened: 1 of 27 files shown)


davidplowman:
@naushir
### DO NOT MERGE!!!

Yes, it's the big shared_ptr change. I think this is one set of changes to think about and try out for a while! It seems like it all works, though I have just once seen a crash using the negate stage - though I can't reproduce it now and don't really know if it was related to any of this work or not.

PR #103, opened 2021-10-01
actions: Add imx477 run test.

naushir:


naushir:
@davidplowman if you hit the merge button, we can have imx477 tests running as well.

PR #104, opened 2021-10-04
test: Add post-processing tests

davidplowman:
Runs a small selection of post-processing tests. Warnings are printed
if OpenCV, TFLite or the model used by TFLite are not found, but the
tests are not marked as failures.


davidplowman:
@naushir All seems to be working as intended now, including the new tests.

PR #109, opened 2021-10-04
core: Initialise control list correctly as "controls::controls"

davidplowman:
This is necessary to stop the IPA process from aborting if it is
running in "isolation" mode.


davidplowman:
@naushir So this small patch is all we need at the libcamera-apps end for our IPAs to work in isolation mode, and in fact this will allow anything that doesn't mode-switch to work. For mode switches, there's a libcamera patch that will have to be merged as well.

PR #110, opened 2021-10-05
test: Add a test that runs with the IPAs in "isolation" mode

davidplowman:
This checks that we aren't sending any ControlLists to libcamera that
could potentially cause problems.


davidplowman:
@naushir Of course we probably want to delay merging this until we've got the necessary patches into libcamera - currently the new test is failing as we would expect!

PR #113, opened 2021-10-06
build: Make support for different types of preview window optional

davidplowman:
cmake will now accept the arguments ENABLE_DRM, ENABLE_X11 and
ENABLE_QT to enable the DRM, X11/GLES and QT preview windows (all
default to 1, i.e. enable).

There are a couple of small tidies, plus the images library is now not
linked in every application but only those that need it.


davidplowman:
@naushir As far as I can tell this all works. Would it be easy to add another build test that runs cmake with `-DENABLE_DRM=0 -DENABLE_X11=0 -DENABLE_QT=0 -DENABLE_OPENCV=0 -DENABLE_TFLITE=0`? I don't think there's any need to run any further tests, but just checking that it builds might be reassuring?

Issue #122, opened 2021-10-11
Wrong version reported by --version  option


dwrobel:
The --version option report wrong version:
```
$ libcamera-jpeg --version
libcamera-apps build: 81925eccaf76 11-10-2021 (17:11:09)
libcamera build: v0.0.0+1-c1d703c3
```
The correct versions are:
```
$ rpm -qv libcamera-apps
libcamera-apps-0.0.0-0.7.20211011git0ddf74f.fc34.armv7hl
$ rpm -qv libcamera
libcamera-0.0.0-0.11.20211009gitfa7a95b.fc34.armv7hl
```
which translates to:
- libcamera-apps@https://github.com/raspberrypi/libcamera-apps/commit/0ddf74f1a0705961bc281eb4da914ec20ce6a671
- https://github.com/libcamera-org/libcamera/commit/fa7a95bf32c6a8ee87b67159d4238cbc8e6dc3a9
respectively.

The problem comes from the fact that the RPM build system _(but it applies to many other build systems)_ fetches tarball based on the gitsha from github and such a tarball is then converted to git repository where other local patches can be easilly applied. 

Please note, that the git repository created by the build system does not have any references to the original github repository _(tarball does not convery git repository - see Source0 example below)_. Not to mention that when no patches are to be applied then the git repository might not be created out of tarball.

So, please consider to use other more reliable system for retrieving software version as the case that original git repository will be used is very uncommon for most linux distributions.

Example what I currently use for building libcamera-apps _(source tarball + patch)_:
```
$ spectool -l libcamera-apps.spec 
Source0: https://github.com/raspberrypi/libcamera-apps/archive/0ddf74f1a0705961bc281eb4da914ec20ce6a671/libcamera-apps-0ddf74f1a0705961bc281eb4da914ec20ce6a671.tar.gz
Patch0: https://github.com/raspberrypi/libcamera-apps/pull/121.patch#/libcamera-apps-Do-not-override-CXXFLAGS-from-external-build-system.patch
```

PR #124, opened 2021-10-12
core: Allow version sha to be passed in at the cmake command line

naushir:
For distribution package builds, the caller may pass a -DVERSION_SHA=... command
line argument to the cmake command to override the version script from getting
it automatically.  This is needed as the packaging process may not use the same
(or indeed any) git tree to do the build.

This change also add an '-auto' tag to builds that auto generate the sha version
string.


naushir:
@davidplowman @XECDesign @dwrobel

You can now run the following:
```
cmake .. -DVERSION_SHA=<...>
```
to manually put the git sha into the version string.  This ought to address https://github.com/raspberrypi/libcamera-apps/issues/122 for libcamera-apps.  The libcamera libraries should really have a similar mechanism.  Until they do, the version string will return 0.0.

PR #130, opened 2021-10-21
build: Add cpack support for src/bin distribution

naushir:
The new make target "package_source" will now save the upstream git commit SHA
into a file that is saved in the source tarball generated.  This SHA will then
be used by the version.py script when building from the tarball, and preserving
the upstream SHA versioning.

This removes the usage of the -DVERSION_SHA=... command line argument.

As a drive-by change, rename the -auto flag to -intree.


naushir:
@davidplowman this is needed for the new version changes.

PR #135, opened 2021-11-04
jpeg, mjpeg: Minor improvements to encoding

davidplowman:
Some slight tidying, and in particular we don't actually need to copy
any pixels beyond the last full block of 16 rows, we can just re-use
pointers to the last row.


davidplowman:
@naushir Tests pass... so it must be good!

Issue #141, opened 2021-11-14
App stop whit multiple img


ml50us:
I need to use -s or -k option to take multiple pict, as i do with raspistill:

libcamera-still -t 0 -k -n -o img%05d.jpg 

But app stop after the first img. It works with timelaps option.
Tks

davidplowman:
Ah, I hadn't realised that was the expect behaviour, but it makes sense. Thanks for pointing it out!

ml50us:
Thanks a lot, its really important for my project. Is it possible to go back to raspistill with bullseye?

davidplowman:
Hi, check out https://forums.raspberrypi.com/viewtopic.php?t=323390 or you can of course stay with Buster.

In general image quality and control of the camera should be better with the new stack, but of course if there are particular issues preventing people from using it (other than ones we already know about!) then we'd appreciate hearing about it.

ml50us:
Perfect! Thanks for your time and for  this great project.

davidplowman:
This fix has been merged now. You can get it either by building from source, or we'll be updating the apt packages in the coming days.

naushir:
Closing, as this has now been fixed.  If there are still any further issue, feel free to re-open.

ml50us:
it works now with keypress -k, but with signal -s , it starts well for the first pict and then the process never stop. 

libcamera-still --version
libcamera-apps build: 6145daf735fc-intree 27-11-2021 (07:06:36)
libcamera build: v0.0.0+3156-f4070274

PR #143, opened 2021-11-14
libcamera-still: In keypress/signal mode return to preview after capture

davidplowman:
The intended behaviour is that you can do one capture after
another. The mechanism for quitting is unchanged (x+Enter or USR2).


davidplowman:
@naushir Fixes https://github.com/raspberrypi/libcamera-apps/issues/141

PR #145, opened 2021-11-17
core: Allow camera selection through the command line.

naushir:
Add a --list-cameras option to list the available cameras that could be
selected for use along with their index values.

Add a --camera <index> option to use the camera given by the index value.

We need an #undef Status just after the X11 library includes as the inclusion
of new headers in this change require it, otherwise we end up with a compile
error.


naushir:
Of course, this cannot be fully utilised until the pipeline handler changes are merged.  However, they do no harm going in early.

PR #146, opened 2021-11-18
export MJPEG frames

christianrauch:
~~I was looking into running the libcamera apps on a Linux desktop and added some useful changes to debug format issues.~~

This PR now implements exporting frames in MJPEG format. Since a MJPEP frame is just a JPEG frame, this will basically:
- write out the raw MJPEG data and skip any EXIF data or thumbnails
- ignore any quality or other compression parameters

Via this PR, the reading and writing of images can be done with integrated laptop cameras on a Linux desktop or with externally attached cameras that provide data in raw JPEG format.

I recommend using an external library to decode/encode between different image formats, if these apps are extended to other formats. One popular library is OpenCV with `cv::imdecode` that automatically determines the data format, and `cv::imencode` which encodes any decoded image into a given format.

### DIFF
--- a/core/libcamera_app.cpp
+++ b/core/libcamera_app.cpp
@@ -12,7 +12,7 @@
 #include "core/options.hpp"
 
 LibcameraApp::LibcameraApp(std::unique_ptr<Options> opts)
-	: options_(std::move(opts)), preview_thread_(&LibcameraApp::previewThread, this), controls_(controls::controls),
+	: options_(std::move(opts)), controls_(controls::controls),
 	  post_processor_(this)
 {
 	if (!options_)
@@ -26,7 +26,8 @@ LibcameraApp::~LibcameraApp()
 		preview_abort_ = true;
 		preview_cond_var_.notify_one();
 	}
-	preview_thread_.join();
+	if (preview_thread_.joinable())
+		preview_thread_.join();
 	if (options_->verbose && !options_->help)
 		std::cerr << "Closing Libcamera application"
 				  << "(frames displayed " << preview_frames_displayed_ << ", dropped " << preview_frames_dropped_ << ")"
@@ -46,6 +47,9 @@ void LibcameraApp::OpenCamera()
 	// Make a preview window.
 	preview_ = std::unique_ptr<Preview>(make_preview(options_.get()));
 	preview_->SetDoneCallback(std::bind(&LibcameraApp::previewDoneCallback, this, std::placeholders::_1));
+
+	if (!options_->nopreview)
+		preview_thread_ = std::thread(&LibcameraApp::previewThread, this);
 
 	if (options_->verbose)
 		std::cerr << "Opening camera..." << std::endl;
--- a/image/jpeg.cpp
+++ b/image/jpeg.cpp
@@ -533,6 +533,8 @@ static void create_exif_data(PixelFormat const &pixel_format, std::vector<libcam
 							options->thumb_height, q, 0, thumb_buffer, thumb_len);
 				break;
 			default:
+				thumb_buffer = nullptr;
+				thumb_len = 0;
 				throw std::runtime_error("unsupported format in JPEG encode: "+pixel_format.toString());
 			}
 
@@ -543,7 +545,7 @@ static void create_exif_data(PixelFormat const &pixel_format, std::vector<libcam
 		}
 		if (options->verbose)
 			std::cerr << "Thumbnail size " << thumb_len << std::endl;
-		if (q <= 0)
+		if (q <= 0 && thumb_buffer != nullptr)
 			throw std::runtime_error("failed to make acceptable thumbnail");
 
 		// Now fill in the correct offsets and length.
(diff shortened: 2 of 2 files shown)


naushir:
@christianrauch MJPEG support is already included in libcamera-apps.  You can capture a stream with the following command:

```
libcamera-vid -t 10000 --codec mjpeg -o test.mjpeg
```

For further information, see [here](https://www.raspberrypi.com/documentation/accessories/camera.html#libcamera-vid).

If you feel this PR adds more that what we already have, please do let us know, otherwise feel free to close this.

christianrauch:
> If you feel this PR adds more that what we already have, please do let us know

Well, yes. If you read through the PR you will notice that this is about exporting JPEG image frames, not videos. With this PR you can do `libcamera-still --nopreview -o image.jpg` to export a JPEG image. You couldn't do this before.

> MJPEG support is already included in libcamera-apps

Have you actually tried `libcamera-vid` on a desktop? On my system, the command you provided gives me a corrupted video, probably because of wrong format conversations. Anyway, this PR is about exporting a single JPEG image frame, not a video.

davidplowman:
Hi Christian

I've been reading through the pull request and perhaps we didn't understand what you want to do. It seems you have a camera that is giving you MJPEG frames and you want to save these to disk, is that right?

For stills capture, forcing an encoded buffer containing JPEG data through `jpeg_save` feels a bit forced to me. I think I'd rather create a "null" still image codec that simply writes the entire buffer to disk. So there'd be a new but very trivial `null.cpp` file in the image folder, and a new `null` encoder option. Does that sound like a way forward?

In the case of video, I'm thinking that `libcamera-raw` might do exactly the right thing already. Whatever comes directly from the camera gets written to disk. You can write them to a single file, or break the file into pieces, even frames.

To answer your other question: no, we don't run these apps on desktop computers. We develop them for use on the Raspberry Pis to work with our supported cameras. However, we're supportive of making these apps useful on other platforms, but the burden of testing and maintaining them on those platforms is not one we ourselves are going to undertake - so 3rd party efforts would be appreciated there.

christianrauch:
> It seems you have a camera that is giving you MJPEG frames and you want to save these to disk, is that right?

Yes, that is correct.

> I think I'd rather create a "null" still image codec that simply writes the entire buffer to disk.

Would this be useful outside of the MJPEG -> JPEG conversion?

> Whatever comes directly from the camera gets written to disk. You can write them to a single file, or break the file into pieces, even frames.

This might be true for the raw output in `libcamera-raw`, but for `libcamera-vid` I would expect that the video gets exported into some readable video format. When I used `libcamera-vid --codec mjpeg -o test.mjpeg` I get a video that looks like it is missing some conversion. I am saying this because I experimented with the code and removed the `Preview windows only support YUV420` exception and the output in the preview window had the same pattern.

I am able to export an MJPEG video with `./libcamera-vid --nopreview --codec=yuv420 --output bla.mjpeg` thought. But while the frames in the video seem to be correct, the timing information seems to be missing as the video just plays as fast as possible.

davidplowman:
> ...
> > I think I'd rather create a "null" still image codec that simply writes the entire buffer to disk.
> 
> Would this be useful outside of the MJPEG -> JPEG conversion?

I think it sounds quite good to have an option that simply writes out whatever the camera spits out at us. But as I said, I can't quite bring myself to take the `jpeg_save` function (which accepts an uncompressed image and outputs a JPEG) to accept an image that is already a JPEG!

> > Whatever comes directly from the camera gets written to disk. You can write them to a single file, or break the file into pieces, even frames.
> 
> This might be true for the raw output in `libcamera-raw`, but for `libcamera-vid` I would expect that the video gets exported into some readable video format. When I used `libcamera-vid --codec mjpeg -o test.mjpeg` I get a video that looks like it is missing some conversion. I am saying this because I experimented with the code and removed the `Preview windows only support YUV420` exception and the output in the preview window had the same pattern.

I'm not quite sure what you tried here. You passed your JPEG buffers that come straight from the camera to the MJPEG encoder? Was that then trying to JPEG them again for a second time?

I think there's a general point here that we're not trying to add support for all different kinds of video output formats. For example, we're not going add code to package stuff up as mp4 files, mkv files or whatever. There's potentially a huge amount of work and in any case, there are a number of big projects that do this sort of thing - vlc, ffmpeg, gstreamer. One solution that people have used successfully is to get `libcamera-vid` to output to stdout and then pipe this into one of these other frameworks. (Though if you have a fairly simple webcam type of device, maybe you don't need libcamera at all?)

> I am able to export an MJPEG video with `./libcamera-vid --nopreview --codec=yuv420 --output bla.mjpeg` thought. But while the frames in the video seem to be correct, the timing information seems to be missing as the video just plays as fast as possible.

I think this is expected. MJPEG normally/often has no timestamp information and players will render it at (for example) 30fps, or indeed "as fast as possible". MJPEG is not a nice format, but then it's not "really" a proper format at all.

davidplowman:
Hi Christian, I think I'm going to close this as I don't think we're going to merge it in the current form. If you wanted to come back to this, please feel free to do so. Perhaps raising an issue first so that we could make a bit of a plan would be helpful too. Thanks!

PR #150, opened 2021-11-19
libcamera-still: allow output folder to be given for date/timestamp o…

davidplowman:
…ptions

The -o (--output) option can now be used to specify a folder name for
the output file when the --datetime or --timestamp options are used.


davidplowman:
@naush Prompted by https://forums.raspberrypi.com/viewtopic.php?t=324149&sid=0e186b05c58b86e0453c22e2dfdf3f91
It seems like a helpful thing to do.

PR #155, opened 2021-11-23
core: Check we are not running the legacy camera stack

davidplowman:
If we definitely are, give up. Anything else - even if not a Pi at all
- we let through.

### DIFF
--- a/core/libcamera_app.cpp
+++ b/core/libcamera_app.cpp
@@ -11,10 +11,38 @@
 #include "core/libcamera_app.hpp"
 #include "core/options.hpp"
 
+#include <fcntl.h>
+
+#include <sys/ioctl.h>
+
+#include <linux/videodev2.h>
+
+// If we definitely appear to be running the old camera stack, complain and give up.
+// Everything else, Pi or not, we let through.
+
+static void check_camera_stack()
+{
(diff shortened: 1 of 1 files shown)


davidplowman:
@naushir Failure looks to be another timing thing... time to buy those cheapo SSDs!

core/libcamera_app.cpp
```diff
@@ -11,10 +11,38 @@
...
+
+	v4l2_capability caps;
+	int ret = ioctl(fd, VIDIOC_QUERYCAP, &caps);
+	close(fd);
+
+	if (ret < 0 || strcmp((char *)caps.driver, "bm2835 mmal"))
+		return;
+
+	fprintf(stderr, "ERROR: the system appears to be configured for the legacy camera stack\n");
+	exit(-1);
```
naushir, line 37: Should we add instruction (or perhaps a url) to detail how to switch back to libcamera?

davidplowman: Yes, might be a good idea - maybe a link to the part of the documentation page that describes what to do? Of course that would have to be written first... :)

Maybe merge this now, and we can change the message once the documentation has been taken care of?

PR #156, opened 2021-11-26
Add bug report generation tool

naushir:
Run this command as follows:
./bug-report.py -t <timeout> -o <filename.txt> -c "libcamera-still -t 1000 -o bug.jpg"

to gererate a bug report to post on a github issue.


naushir:
@davidplowman 
This this might be finally ready to go in.  The tool will now be installed system-wide with a make install so it does not need to be pulled in manually.  We can also run without the '-c' argument, so generate a report with all the relevant information if the users cannot run the application through the tool.

kbingham:
This is a great idea!
I had been planning something (no where near as good) for libcamera to be able to capture the media graphs and kernel logs.

PR #169, opened 2021-12-02
utils: Add timestamp analysis tool.

naushir:


davidplowman:
Just one little request - it seemed yesterday like outputting the frame number of the biggest gap might be useful, could we add that to this commit?

naushir:
Done!

PR #170, opened 2021-12-02
output: Upgrade circular buffer output for 64-bit systems

davidplowman:
The buffer can now be > 4GB. The buffer size can also be specified on
the command line with the --circular option.

### DIFF
--- a/core/video_options.hpp
+++ b/core/video_options.hpp
@@ -49,8 +49,8 @@ struct VideoOptions : public Options
 			 "Create a new output file every time recording is paused and then resumed")
 			("segment", value<uint32_t>(&segment)->default_value(0),
 			 "Break the recording into files of approximately this many milliseconds")
-			("circular", value<bool>(&circular)->default_value(false)->implicit_value(true),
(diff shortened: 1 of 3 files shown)


core/video_options.hpp
```diff
@@ -49,7 +49,7 @@ struct VideoOptions : public Options
 			 "Create a new output file every time recording is paused and then resumed")
 			("segment", value<uint32_t>(&segment)->default_value(0),
 			 "Break the recording into files of approximately this many milliseconds")
-			("circular", value<bool>(&circular)->default_value(false)->implicit_value(true),
+			("circular", value<size_t>(&circular)->default_value(0)->implicit_value(1<<22),
 			 "Write output to a circular buffer which is saved on exit")
```
naushir, line 53: Should we perhaps use megabyte units here?
Not too fussed either way.

davidplowman: Yes, good suggestion, I think. I've done that now!

naushir:
LGTM, I'll merge once the tests run.
Should get some more test machines!

PR #178, opened 2021-12-03
core: Force still capture to default to full resolution camera mode

davidplowman:
We do this by always requesting a raw stream even though we may not
otherwise use it. This defaults to full resolution, thereby forcing
the selection of that camera mode.

It can be overridden with the --mode option, if desired.


davidplowman:
@naushir This fixes the "poor quality VGA captures" problem reported on the forums. It does change the behaviour of libcamera-still from what it was before, but this better matches raspistill and seems "reasonable"...

PR #184, opened 2021-12-07
Allow circular output to write to stdout

dustinkerstein:
Can be tested with: 
```
libcamera-vid --circular 1000 --frames 60 --framerate 30 --width 1920 --height 1080 --codec yuv420 -o - \
| pv -r > /dev/null
```
And allows for piping to FFmpeg/Gstreamer, with the purpose of holding the encode until all frames have been captured (up to the amount that can fit in circular ram buffer):
```
libcamera-vid --circular 1000 --frames 60 --framerate 30 --width 1920 --height 1080 --codec yuv420 -o - \
| ffmpeg -y -f rawvideo -pix_fmt yuv420p -s:v 1920x1080 -r 30 -i - -c:v h264_v4l2m2m -b:v 25M ~/out.mp4
```


naushir:
Looks good to me.

Issue #163, opened 2021-11-28
[BUG] libcamera-jpeg and libcamera-still SIGSEGV segmentation fault


janvrska:
**Describe the bug**
When I want to capture photo with libcamera-jpeg or libcamera-still  in 90% of attempts it show preview window successfully and then ends with sigsegv segmentation fault and image is not saved. It's interesting that in libcamera-vid saving video works fine.

**To Reproduce**
With backtrace:
```
pi@raspberrypi:~/camera $ gdb libcamera-jpeg
GNU gdb (Raspbian 10.1-1.7) 10.1.90.20210103-git
Copyright (C) 2021 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "arm-linux-gnueabihf".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<https://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from libcamera-jpeg...
(No debugging symbols found in libcamera-jpeg)
(gdb) bt
No stack.
(gdb) run -o test.jpg
Starting program: /usr/local/bin/libcamera-jpeg -o test.jpg
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/arm-linux-gnueabihf/libthread_db.so.1".
[New Thread 0xb232f1c0 (LWP 4797)]
[2:07:59.479617365] [4790]  INFO Camera camera_manager.cpp:293 libcamera v0.0.0+3266-cb1de399
[New Thread 0xaaa961c0 (LWP 4805)]
[2:07:59.486130512] [4805]  WARN CameraSensorProperties camera_sensor_properties.cpp:141 No static properties available for 'imx477'
[2:07:59.486203326] [4805]  WARN CameraSensorProperties camera_sensor_properties.cpp:143 Please consider updating the camera sensor properties database
[New Thread 0xaa0ff1c0 (LWP 4806)]
[New Thread 0xa98fe1c0 (LWP 4807)]
[...]

PR #185, opened 2021-12-08
core: Stop preview thread in Teardown method

davidplowman:
This means it can't access buffers that are being deallocated when the
display is very laggy (such as when using VNC). Correspondingly, the
preview thread is started at the end of the Configure methods.


davidplowman:
@naushir Fix for https://github.com/raspberrypi/libcamera-apps/issues/163, all good now!

PR #190, opened 2021-12-10
preview: Fix more build headaches caused by X11

davidplowman:
This time it's because X11 #defines "None" (which is going to be a
problem when the libcamera colour space changes get merged). This
workaround seems to be adequate.


davidplowman:
@naushir Prophylactic fix for libcamera colour spaces...

PR #191, opened 2021-12-10
Added so_reuseaddr to listening sockets so they can be restarted.

tmorley2000:
Without SO_REUSEADDR listening sockets cannot be re-opened within 2 minutes, this just enables the option so you can restart a listening app without delay.

### DIFF
--- a/output/net_output.cpp
+++ b/output/net_output.cpp
@@ -48,6 +48,10 @@ NetOutput::NetOutput(VideoOptions const *options) : Output(options)
 			server_saddr.sin_addr.s_addr = INADDR_ANY;
 			server_saddr.sin_port = htons(port);
 
+			int enable = 1;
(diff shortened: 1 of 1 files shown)


davidplowman:
Hi, thanks very much for this patch. It looks good to me - would you be OK just to make it pass the style checker? Most of us run it as a pre-commit hook, though in this particular case it's obviously trivial enough to fix. We normally also add a "Signed-off-by" line (check some of the other commits), which is just a kind of declaration that we can distribute the modifications under the existing (BSD 2-Clause) licence. Thanks!

tmorley2000:
I've tidied the style checker complaints.

Signed-off-by: Tim Morley <tim@chaos.org.uk>

davidplowman:
Thanks for doing that. At the risk of being a bit picky, would you mind squashing the two commits together? If we can manage it we prefer to try and keep the commit history as "tidy" as we can. WIth that I think I'm good to merge. Thanks!

tmorley2000:
Ok, I think thats squashed down now!

PR #194, opened 2021-12-13
libcamera-vid: Signals were not being "cancelled" in signal mode

davidplowman:
This copies a previous fix in libcamera-still (fa95bcb), which applies
here too.


davidplowman:
@naushir I should probably have remembered this when I fixed libcamera-still in the first place...

PR #197, opened 2021-12-14
preview: Preview was not resuming properly during timelapse

davidplowman:
A preview fix (9d89e1e) actually broke this. There was a particularly
subtle issue with EGL where it was still only obeying the previous
preview thread which has stopped running.

### DIFF
--- a/core/libcamera_app.cpp
+++ b/core/libcamera_app.cpp
@@ -478,9 +478,6 @@ void LibcameraApp::StopCamera()
 
 	msg_queue_.Clear();
 
-	if (preview_)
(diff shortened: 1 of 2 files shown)


davidplowman:
@naushir Bit of a whoops here. The previous preview fix managed to stop the preview from restarting properly after being stopped, such as when doing a timelapse (though nothing obvious failed or crashed). And EGL is a particular pain because it's fussy about what thread you do things in. Anyway, I think this works now. We probably need to devote more time to improving test coverage... (though not sure how you'd detect this sort of thing other than by watching it).

PR #198, opened 2021-12-14
Colour spaces

davidplowman:
Still WIP - don't merge!

### DIFF
--- a/apps/libcamera_vid.cpp
+++ b/apps/libcamera_vid.cpp
@@ -50,6 +50,14 @@ static int get_key_or_signal(VideoOptions const *options, pollfd p[1])
 	return key;
 }
 
+static int get_colourspace_flags(std::string const &codec)
+{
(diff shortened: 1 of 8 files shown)


davidplowman:
I don't think I'm planning to do any more on this, so I guess that makes it ready for review.

I've rewritten the DRM stuff a bit. Previously I was duplicating the kernel enums so that I could use them directly (they don't seem to be provided to userspace). Now I'm searching for the correct enum values by matching strings - DRM does appear to provide name-to-value lookup tables. But I still don't particularly like it.

naushir:
Should we submit this change now?

davidplowman:
Well, I'm "happy enough" with it, though I can't say I love the messing around with DRM. But it would be good to commit it before we release our updated versions of everything, I think.

naushir:
Done!

PR #214, opened 2021-12-29
core: Rationalise handling of free_requests_ and known_completed_requ…

davidplowman:
…ests_

Firstly, known_completed_requests_ was not being protected by a mutex
(like free_requests_) and should have been. But moreover, we can
dispense with the free_requests_ list completely if the
CompletedRequest contains a pointer to the libcamera Request object,
and this tidies up the whole "queue another Request when the
CompletedRequest is dropped" logic (in LibcameraApp::queueRequest).

In fact we could go further and leave the BufferMap and metadata in
the Request (not copying it to the CompletedRequest), but that would
touch quite a few files so we can save it for another time.

### DIFF
--- a/core/completed_request.hpp
+++ b/core/completed_request.hpp
@@ -18,14 +18,17 @@ struct CompletedRequest
 {
 	using BufferMap = libcamera::Request::BufferMap;
 	using ControlList = libcamera::ControlList;
+	using Request = libcamera::Request;
 
-	CompletedRequest(unsigned int seq, BufferMap const &b, ControlList const &m)
-		: sequence(seq), buffers(b), metadata(m)
+	CompletedRequest(unsigned int seq, Request *r)
+		: sequence(seq), buffers(r->buffers()), metadata(r->metadata()), request(r)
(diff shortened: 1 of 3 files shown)


dustinkerstein:
> In fact we could go further and leave the BufferMap and metadata in
> the Request (not copying it to the CompletedRequest), but that would
> touch quite a few files so we can save it for another time.

Quick note should you eventually decide to go that route - Do you think there would be another easy way to get the awbgain values? I'm currently doing this:

`libcamera::Span<const float> gains = completed_request->metadata.get(libcamera::controls::ColourGains);`

davidplowman:
I doubt that would change much, it would probably look like `completed_request->metadata().get(...);` which is hardly any different (just an extra pair of brackets).

(Wandering slightly off topic...) You could also do this:
```
    FrameInfo info(completed_request.metadata);
```
which gives you the colour gains again, but also quite a few extra fields.

dustinkerstein:
Awesome. Thanks for clarifying.

PR #222, opened 2022-01-10
options: Support "long" exposure mode

davidplowman:
The exposure options did not include this, so it has been added.


davidplowman:
@naushir No idea why I didn't put in this exposure mode...

PR #223, opened 2022-01-13
utils: Require the -c argument for camera-bug-report

naushir:
Without this argument, we do not know what to run.


naushir:
@davidplowman

PR #255, opened 2022-02-21
preview: Clear the preview item queue when we stop the preview

davidplowman:
preview_item_ is effectively a 1-deep queue of frames still to
display. It's best to clear this when the preview is stopped so that
there's no way that anyone might attempt to display it if the preview
restarts (by which time it would be stale).


naushir (commented):
LGTM

PR #256, opened 2022-02-21
Include sstream header

tewarid:
Fix for clang fatal error: implicit instantiation of undefined template 'std::basic_stringstream<char>'

### DIFF
--- a/core/frame_info.hpp
+++ b/core/frame_info.hpp
@@ -6,6 +6,7 @@
  */
 #include <array>
 #include <iomanip>
+#include <sstream>
 #include <string>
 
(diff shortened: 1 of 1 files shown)


naushir:
@tewarid thank you for this patch - and resubmitting the commit with the style checker fix.
Once all the checks have been completed, I'll merge this in.

davidplowman:
Just out of curiosity, do we know why our smoke tests don't pick this up, it does a couple of clang builds, right?

naushir:
I'm suspecting it's down to our clang version on the test systems.  Or possibly this is a new/custom app that does not include the same header definitions?

PR #264, opened 2022-02-26
Timeout Warning & Build Instructions 

jlpoolen:
1) prints to console timeout value, if previously set and causing the termination of the process
2) isolated Build instructions with its own heading.  In text mode, a link at the end of an introductory paragraph can be missed.

This is my second pull request, the previous one failed because I had not correctly run checkstyle.py _after_ committing rather than before.

### DIFF
--- a/apps/libcamera_vid.cpp
+++ b/apps/libcamera_vid.cpp
@@ -97,8 +97,7 @@ static void event_loop(LibcameraEncoder &app)
 		if (timeout || frameout || key == 'x' || key == 'X')
 		{
 		        if (timeout)
-			  std::cout << "Halting: reached timeout of " << options->timeout
-				    << " milleseconds.\n";
+			        std::cout << "Halting: reached timeout of " << options->timeout << " milliseconds.\n";
 			
 			app.StopCamera(); // stop complains if encoder very slow to close
 			app.StopEncoder();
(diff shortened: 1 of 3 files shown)


naushir:
Thank you for your PR.  Happy with the changes themselves, however, would you be able to provide two separate commits:

- One for the build instructions change.
- Second one for the timeout log message.

The merge commits and formatting fixes should not be part of the PR.

apps/libcamera_vid.cpp
```diff
@@ -96,6 +96,9 @@ static void event_loop(LibcameraEncoder &app)
 		bool frameout = options->frames && count >= options->frames;
 		if (timeout || frameout || key == 'x' || key == 'X')
 		{
+		        if (timeout)
+				std::cout << "Halting: reached timeout of " << options->timeout << " milliseconds.\n";
+
 			app.StopCamera(); // stop complains if encoder very slow to close
```
naushir, line 102: This wants to be std::cerr

jlpoolen:
Will do.

apps/libcamera_vid.cpp
jlpoolen: Agreed.

jlpoolen:
Redoing per request to isolate changes and direct to standard error.

PR #267, opened 2022-02-27
Isolated build instructions so link is more prominent in a text mode

jlpoolen:
Since build instructions are maintained within another doc that is linked to, highlighted by added subtitle "Build" so reader in text mode sees that another document should be referenced.


jlpoolen:
I hope I have the commits isolated, but within 1 pull, is satisfactory.  I'm not familiar with git other than to pull and push, so this is my first time working in a more structured setting.  Thank you for your patience and sorry my efforts may be consuming more time than if someone knowledgeable had incorporated the changes.

davidplowman:
Thanks!

PR #282, opened 2022-03-30
Do not advertise USB webcams, they are not supported in libcamera-apps.

naushir:


6by9:
Do we ignore USB webcams, or only support devices under /base/soc?
Blacklist or whitelist?

davidplowman:
LGTM.

PR #285, opened 2022-04-07
Dng fixes

davidplowman:
Couple of commits here. The first should be uncontroversial, it's just adding the DateTimeOriginal EXIF tag.

The second relates to some strange libtiff behaviour, or at least behaviour I don't understand. This commit adds a workaround for the problem, though it would be helpful if anyone has a better understanding of what is going on. So it probably needs some thinking about. Hence a "draft" PR for now.

### DIFF
--- a/image/dng.cpp
+++ b/image/dng.cpp
@@ -304,6 +304,14 @@ void dng_save(std::vector<libcamera::Span<uint8_t>> const &mem, StreamInfo const
 		// Create a separate IFD just for the EXIF tags. Why we couldn't simply have
 		// DNG tags for these, which would have made life so much easier, I have no idea.
 		TIFFCreateEXIFDirectory(tif);
+
+		time_t t;
+		time(&t);
+		struct tm *time_info = localtime(&t);
+		char time_str[32];
+		strftime(time_str, 32, "%Y:%m:%d %H:%M:%S", time_info);
+		TIFFSetField(tif, EXIFTAG_DATETIMEORIGINAL, time_str);
+
 		TIFFSetField(tif, EXIFTAG_ISOSPEEDRATINGS, 1, &iso);
 		TIFFSetField(tif, EXIFTAG_EXPOSURETIME, exp_time);
 
--- a/image/dng.cpp
+++ b/image/dng.cpp
@@ -325,6 +325,14 @@ void dng_save(std::vector<libcamera::Span<uint8_t>> const &mem, StreamInfo const
 		TIFFSetField(tif, TIFFTAG_EXIFIFD, offset_exififd);
 		TIFFWriteDirectory(tif);
 
+		// For reasons unknown, the last sub-IFD that we make seems to reappear at the
+		// end of the file as IDF1, and some tools (exiftool for example) are prone to
+		// complain about it. As far as I can see the code above is doing the correct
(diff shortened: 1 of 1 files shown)


davidplowman:
Interestingly, I find that with this change, DarkTable will now load our DNGs which it wouldn't previously (though other packages did). So maybe this fix is even better than I realised!

davidplowman:
@naushir It seems to work for me, and there do appear to be real benefits, so I reckon we should give it a try...!

image/dng.cpp
```diff
@@ -304,6 +304,14 @@ void dng_save(std::vector<libcamera::Span<uint8_t>> const &mem, StreamInfo const
 		// Create a separate IFD just for the EXIF tags. Why we couldn't simply have
 		// DNG tags for these, which would have made life so much easier, I have no idea.
```
kmilos, line 305: But you do - it's called [TIFF/EP](https://en.wikipedia.org/wiki/TIFF/EP) (which DNG extends if you've read the spec carefully, as do other raw containers like NEF) an these tags go straight into IFD0 an replicate most of Exif tags. Everything is in these two specs:

http://www.barrypearson.co.uk/top2009/downloads/TAG2000-22_DIS12234-2.pdf 
https://helpx.adobe.com/content/dam/help/en/photoshop/pdf/dng_spec_1_6_0_0.pdf

So typically DNGs don't have an ExifIFD unless there are really some other tags needed that don't exist in TIFF/EP+DNG. Usually you just have IFD0 w/ preview and all the metadata, and SubIFD w/ the raw image.

davidplowman: Hi, thanks for that information. In fact I do seem to remember trying to get it to accept those tags in IFD0 but I never got it to work. I always got "unknown tag" errors, unless I was putting them in the EXIFID. Do you know if libtiff supports this feature? Perhaps libtiff isn't the best library to use, do you know of any alternatives? Thanks again!

kmilos: Ah, haven't tried that in a while, might be a problem indeed. Perhaps that's best raised w/ libtiff as an unnecessary limitation.

As a workaround, you could close the file w/o these "Exif" specific tags w/ libtiff, then add them via `exiv2` library API to the Exif.Image directory (i.e. IFD0), rather than Exif.Photo (ExifIFD). (There is also `libexif` which might be lighter but don't know if it'll do the job...)

davidplowman: To be honest, I've found writing DNG files in C/C++ to be unreasonably difficult. Given that dcraw, RawTherapee, DarkTable and the Python libraries we use in our own software all seem to accept them, I'm not very keen to spend more time on it. Moving forwards we're going to be making ever more use of Picamera2 which supports DNG using the PiDNG Python library.

Entropy512: @davidplowman The PiDNG description at https://github.com/schoolpost/PiDNG kind of implies it should have generic DNG support, any reason it was named PiDNG as opposed to PyDNG?

Thanks for mentioning that, it looks like it might be the solution to some of my headaches of writing DNGs from Python after reading raw data with rawpy.

davidplowman: I think the PiDNG was originally for the Pi but it has expanded since. You'd have to check with the author who has no affiliation with Raspberry Pi but has very generously been happy to help us make use of his work.

PR #288, opened 2022-04-12
Use controls::SensorTimestamp for timestamps if possible.

naushir:
Framebuffer timestamps account for SW overheads and pipelining, so might be a
bit glitchy.


davidplowman:
We've decided that we like the indentation as it is so we're going to ignore these specific style checker complaints.

PR #294, opened 2022-05-18
annotate_cv_stage: Allow date/time format directives (as per strftime)

davidplowman:
raspivid used to do this, so we should too.


davidplowman:
@naushir Don't know if you spotted this little one...

PR #296, opened 2022-05-23
Autofocus patches from ArduCam 

supertobi:


supertobi:
I'm not sure if it is useful for you at all, but it would make live easier for all users with a ArduCam autofocus camera.

naushir:
Unfortunately we cannot merge these into the tree.  The AF controls are not yet supported by libcamera.  We are, however, in the process of getting AF control merged to the core libcamera libraries.

Arducam have changes to our libcamera pipeline handler that are not part of the mainline release either, so these changes cannot be tested by us.

PR #298, opened 2022-05-30
Add libav encoder

naushir:


naushir:
This is just a draft for review.  I intend on squashing everything into a single commit once we are ready to merge this.

encoder/libav_encoder.hpp
```diff
@@ -0,0 +1,71 @@
...
+
+	std::queue<AVFrame *> frame_queue_;
+	std::mutex video_mutex_;
+	std::mutex output_mutex_;
+	std::condition_variable video_cv_;
+	std::thread video_thread_;
+	std::thread audio_thread_;
+
+	// The ordering in the enum below must not change!
+	enum Context { Video, AudioOut, AudioIn };
```
davidplowman, line 65: I don't suppose we could make this more explicit or less fragile with a `Video = 0, AudioOut = 1, AudioIn = 2` kind of thing?

output/output.cpp
```diff
@@ -70,6 +70,9 @@ void Output::outputBuffer(void *mem, size_t size, int64_t timestamp_us, uint32_t
 
 Output *Output::Create(VideoOptions const *options)
 {
+	if (options->codec == "libav")
+		return new Output(options);
+
```
davidplowman, line 75: Hehe. An output that does nothing at all. I love it!

core/video_options.hpp
```diff
@@ -54,6 +58,25 @@ struct VideoOptions : public Options
...
+			 "To list available codecs, run  the \"ffmpeg -codecs\" command.")
+			("audio-device", value<std::string>(&audio_device)->default_value("default"),
+			 "Audio device to record from. To list the available devices, use the following command:\n"
+			 "pactl list | grep -A2 'Source #' | grep 'Name: '")
+			("audio-bitrate", value<uint32_t>(&audio_bitrate)->default_value(32768),
+			 "Set the audio bitrate for encoding, in bits/second.")
+			("av-sync", value<int32_t>(&av_sync)->default_value(0),
+			 "Add a time offset (in microseconds) to the audio stream, relative to the video stream. "
+			 "The offset value can be either positive or negative.")
+#endif
```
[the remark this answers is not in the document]
naushir: The audio bitrate is specified in b/s, or have I misinterpreted your comment?

Currently the samplerate matches the input.  There ought to be no reason not to specify this on the command line, but I did not have a chance to fully test my resampling code, so I thought I'd not allow setting this until it was fully proven.

encoder/CMakeLists.txt
```diff
@@ -2,8 +2,28 @@ cmake_minimum_required(VERSION 3.6)
...
+    include_directories(${LIBAV_INCLUDE_DIRS})
+    set(SRC ${SRC} libav_encoder.cpp)
+    set(TARGET_LIBS ${TARGET_LIBS} ${LIBAV_LIBRARIES})
+    set(LIBAV_PRESENT 1)
+    message(STATUS "    libraries: ${LIBAV_LIBRARIES}")
+endif()
+
+add_library(encoders ${SRC})
+target_link_libraries(encoders ${TARGET_LIBS})
+target_compile_definitions(encoders PUBLIC LIBAV_PRESENT=${LIBAV_PRESENT})
```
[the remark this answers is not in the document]
naushir: Our apt packages do include libav (for ffmpeg), so this will be built by default.  I wanted to make sure I did not trip over anyone else who might not have libav for whatever reason and wanted to do a custom build.

encoder/libav_encoder.hpp
naushir: Yes, I can do that!

output/output.cpp
naushir: If for some reason a user wanted to output a .pts file when using libav, this would allow it. Not sure how useful that may be, given the timestamps are also encoded in the container...

encoder/libav_encoder.cpp
```diff
@@ -0,0 +1,548 @@
...
+
+	codec_ctx_[Video] = avcodec_alloc_context3(codec);
+	if (!codec_ctx_[Video])
+		throw std::runtime_error("libav: Cannot allocate video context");
+
+	codec_ctx_[Video]->width = info.width;
+	codec_ctx_[Video]->height = info.height;
+	// usec timebase
+	codec_ctx_[Video]->time_base = { 1, 1000 * 1000 };
+	codec_ctx_[Video]->framerate = { (int)(options->framerate * 1000), 1000 };
```
davidplowman, line 37: I assume these are just rational numbers. I wonder if there's some syntactic sugar that would make this more obvious in some way, but I'm probably just fussing because, frankly, most of this is going over my head!!

encoder/libav_encoder.cpp
```diff
@@ -0,0 +1,548 @@
...
+	codec_ctx_[Video]->pix_fmt = AV_PIX_FMT_DRM_PRIME;
+	codec_ctx_[Video]->sw_pix_fmt = AV_PIX_FMT_YUV420P;
+
+	if (info.colour_space)
+	{
+		using libcamera::ColorSpace;
+
+		static const std::map<ColorSpace::Primaries, AVColorPrimaries> pri_map = {
+			{ ColorSpace::Primaries::Raw, AVCOL_PRI_UNSPECIFIED },
```
davidplowman, line 46: Just curious, could this case ever happen?

encoder/libav_encoder.cpp
```diff
@@ -0,0 +1,548 @@
...
+		if (it_tf != tf_map.end())
+			codec_ctx_[Video]->color_trc = it_tf->second;
+
+		auto it_cs = cs_map.find(info.colour_space->ycbcrEncoding);
+		if (it_cs != cs_map.end())
+			codec_ctx_[Video]->colorspace = it_cs->second;
+
+		codec_ctx_[Video]->color_range =
+			info.colour_space->range == ColorSpace::Range::Full ? AVCOL_RANGE_JPEG : AVCOL_RANGE_MPEG;
+	}
```
davidplowman, line 79: Do we need to worry about the possibility of failure cases in these lookups here? I agree I can't see why it would happen...

encoder/libav_encoder.cpp
```diff
@@ -0,0 +1,548 @@
...
+		codec_ctx_[Video]->profile = it->second;
+	}
+
+	if (!options->level.empty())
+	{
+	}
+
+	if (options->inline_headers)
+	{
+	}
```
davidplowman, line 106: Interesting, so there's no option for repeating the headers in every I frame. I guess we don't care? I suppose if you were streaming elementary H.264 to a socket then you'd have no reason to crank up the whole of FFmpeg, and otherwise it would take care of that kind of thing (headers quite often go out-of-band after all, don't they).

encoder/libav_encoder.cpp
naushir: Yes, they are rational numbers.  Essentially, I want to match the units of the frame timestamps for simplicity.

encoder/libav_encoder.cpp
naushir: Should not... I can remove this.

encoder/libav_encoder.cpp
naushir: I'll put something in to be completely safe.

encoder/libav_encoder.cpp
naushir: Yes, this is more a limitation of the libav interface to the v4lm2m codec (as far as I can tell).  But luckily libav does manage to do the right thing internally when outputting to files/containers vs streaming.

6by9: At the V4L2 level it does set V4L2_CID_MPEG_VIDEO_HEADER_MODE to HEADER_MODE_SEPARATE to get the headers in a separate buffer to the I-frame, but it never sets V4L2_CID_MPEG_VIDEO_REPEAT_SEQ_HEADER.

It could repeat the headers within the FFmpeg code, but it doesn't ask V4L2 to do it.

encoder/libav_encoder.cpp
```diff
@@ -0,0 +1,548 @@
...
+	if (!frame)
+		throw std::runtime_error("could not allocate AVFrame");
+
+	if (!video_start_ts_)
+		video_start_ts_ = timestamp_us;
+
+	frame->format = AV_PIX_FMT_DRM_PRIME;
+	frame->width = info.width;
+	frame->height = info.height;
+	frame->linesize[0] = info.stride;
```
davidplowman, line 268: It assumes U and V planes to have half the stride of Y? (I know it's true, of course)

davidplowman: Ah, I see they get put in the `desc` object below!

encoder/libav_encoder.cpp
```diff
@@ -0,0 +1,548 @@
...
+		}
+		else if (ret < 0)
+			throw std::runtime_error("libav error receiving packet: " + std::to_string(ret));
+
+		// Initialise the ouput mux on the first received video packet, as we may need
+		// to copy global header data from the encoder.
+		if (stream_id == Video && !init_output_)
+		{
+			initOutput();
+			init_output_ = true;
```
davidplowman, line 357: `init_output_ = true` sounds a bit like the output still needs to be init-ed, but I realise I'm splitting hairs and probably being very annoying.

naushir: I'll rename to output_ready_ :)

encoder/libav_encoder.cpp
```diff
@@ -0,0 +1,548 @@
...
+
+	if (options_->verbose)
+		std::cerr << "libav closed" << std::endl;
+}
+
+void LibAvEncoder::EncodeBuffer(int fd, size_t size, void *mem, StreamInfo const &info, int64_t timestamp_us)
+{
+	AVFrame *frame = av_frame_alloc();
+	if (!frame)
+		throw std::runtime_error("could not allocate AVFrame");
```
davidplowman, line 260: Might be nice if all the errors had a common preamble (I think pretty much everywhere else I see `libav:`), though I agree this one is pretty obvious.

naushir: Will do!

encoder/libav_encoder.cpp
```diff
@@ -0,0 +1,548 @@
...
+
+		pkt->stream_index = stream_id;
+		pkt->pos = -1;
+		pkt->duration = 0;
+
+		// Rescale from the codec timebase to the stream timebase.
+		av_packet_rescale_ts(pkt, codec_ctx_[stream_id]->time_base, out_fmt_ctx_->streams[stream_id]->time_base);
+
+		std::scoped_lock<std::mutex> lock(output_mutex_);
+		// pkt is now blank (av_interleaved_write_frame() takes ownership of
```
davidplowman, line 368: Am I right in thinking this mutex only ever gets grabbed once? That's feeling a bit weird to me...

davidplowman: Ah, or do we have two threads calling this function?

davidplowman:
@naushir Are you good for me to push the big green button?

naushir:
I see no reason why not :)

PR #301, opened 2022-06-16
libav: Update the audio-codec command line option to be consistent.

naushir:


naushir:
@davidplowman just a small change to be consistent.

Issue #303, opened 2022-06-19
[BUG] Memory leak in DRM preview


asmirnou:
**Describe the bug**

Raspberry Pi 4 with Lite OS and HQ camera, HDMI monitor plugged in.
libcamera-still in timelapse mode having captured 9-10 images fails to allocate more frame buffers. 
With ` --nopreview` option images are captured continuously. 

```
libcamera-still --timeout 0 --timelapse 2000 -o test.jpg
```

```
...
Still capture image received
[0:20:03.496245571] [3271]  INFO Camera camera.cpp:1029 configuring streams: (0) 2028x1520-YUV420
[0:20:03.516958039] [3272]  INFO RPI raspberrypi.cpp:760 Sensor: /base/soc/i2c0mux/i2c@1/imx477@1a - Selected sensor format: 2028x1520-SBGGR12_1X12 - Selected unicam format: 2028x1520-pBCC
[0:20:04.303260783] [3271]  INFO Camera camera.cpp:1029 configuring streams: (0) 4056x3040-YUV420 (1) 4056x3040-SBGGR12_CSI2P
[0:20:04.305372158] [3272]  INFO RPI raspberrypi.cpp:760 Sensor: /base/soc/i2c0mux/i2c@1/imx477@1a - Selected sensor format: 4056x3040-SBGGR12_1X12 - Selected unicam format: 4056x3040-pBCC
Still capture image received
[0:20:05.695079489] [3271]  INFO Camera camera.cpp:1029 configuring streams: (0) 2028x1520-YUV420
[0:20:05.716344063] [3272]  INFO RPI raspberrypi.cpp:760 Sensor: /base/soc/i2c0mux/i2c@1/imx477@1a - Selected sensor format: 2028x1520-SBGGR12_1X12 - Selected unicam format: 2028x1520-pBCC
[0:20:06.503059414] [3271]  INFO Camera camera.cpp:1029 configuring streams: (0) 4056x3040-YUV420 (1) 4056x3040-SBGGR12_CSI2P
[0:20:06.505150535] [3272]  INFO RPI raspberrypi.cpp:760 Sensor: /base/soc/i2c0mux/i2c@1/imx477@1a - Selected sensor format: 4056x3040-SBGGR12_1X12 - Selected unicam format: 4056x3040-pBCC
[...]

PR #304, opened 2022-06-21
previews: Fix memory leak in DRM preview

davidplowman:
"bo_handles" need to be closed, and although I can't find a "regular"
API call there does appear to be a DRM ioctl that does this.


davidplowman:
This seems to fix https://github.com/raspberrypi/libcamera-apps/issues/303 for me. I think it's the correct solution, though I can't claim any great expertise on DRM...

naushir:
Good enough for me.

PR #305, opened 2022-06-22
core: Display available controls with the verbose flag.

naushir:
It might be useful for the user to know the sensor control limits.


naushir:
@davidplowman

PR #308, opened 2022-06-27
libcamera-still: Allow --immediate and keypress/signal to work together

naushir:
These command line arguments set together would not do the expected thing.
Fix this by having a wait loop after configuring stills mode and wait for the
signal/keypress before taking the capture.

### DIFF
--- a/apps/libcamera_still.cpp
+++ b/apps/libcamera_still.cpp
@@ -4,7 +4,7 @@
  *
  * libcamera_still.cpp - libcamera stills capture app.
  */
-
+#include <chrono>
 #include <poll.h>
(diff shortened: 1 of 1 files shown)


naushir:
@davidplowman - this one is ready as well.

davidplowman:
I guess this is all ok, but I don't love it. I can't escape the feeling the users would really want it to be running in stills capture mode up front, so that the delay when the capture signal arrives is minimised. But at that point we're shoehorning yet more convoluted modes of operation into something that's already too way too fiddly. So maybe we just bin that idea and tell folks to go and write some Picamera2 instead. Sigh.

PR #318, opened 2022-07-15
output: Circular Buffer only save timestamps for used frames

will-v-pi:
When using a circular buffer output and saving the timestamps to a file,
only save the timestamps for the frames which have been saved, rather
than saving all the timestamps even for frames no longer in the buffer.

Also allow --flush to flush the pts values to a file.

### DIFF
--- a/output/output.cpp
+++ b/output/output.cpp
@@ -14,7 +14,7 @@
 #include "output.hpp"
 
 Output::Output(VideoOptions const *options)
-	: options_(options), state_(WAITING_KEYFRAME), fp_timestamps_(nullptr), time_offset_(0), last_timestamp_(0)
+	: options_(options), fp_timestamps_(nullptr), state_(WAITING_KEYFRAME), time_offset_(0), last_timestamp_(0)
 {
 	if (!options->save_pts.empty())
 	{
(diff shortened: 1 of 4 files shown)


output/output.hpp
```diff
@@ -31,7 +31,9 @@ class Output
 		FLAG_RESTART = 2
 	};
 	virtual void outputBuffer(void *mem, size_t size, int64_t timestamp_us, uint32_t flags);
+	virtual void timestampReady();
 	VideoOptions const *options_;
+	FILE *fp_timestamps_;
```
naushir, line 37: One thing I might change here is to make timestampReady() and fp_timestamps_ protected members rather than public.  That way, the derived classes can still access them, but the application cannot.

output/output.cpp
```diff
@@ -60,7 +60,16 @@ void Output::OutputReady(void *mem, size_t size, int64_t timestamp_us, bool keyf
...
+		timestampReady();
+	}
+}
+
+void Output::timestampReady()
+{
+	fprintf(fp_timestamps_, "%" PRId64 ".%03" PRId64 "\n", last_timestamp_ / 1000, last_timestamp_ % 1000);
+	if (options_->flush)
+		fflush(fp_timestamps_);
 }
```
naushir, line 73: What do you think about passing a timestamp value as a parameter into Output::timestampReady()?

This will allow us CircularOutput::~CircularOutput() to call into Output::timestampReady() with the timestamp values passed in, and avoid duplicating this code.

PR #320, opened 2022-07-18
Add verbosity levels

will-v-pi:
Add 3 separate verbosity levels, with 0 having no output, 1 having the
same effect as not setting -v previously, and 2 having the same effect
as setting -v previously.

Maintains same behaviour as setting/not setting -v previously, and just
adds level 0, which can be accessed by using -v 0.

Example uses:
Level 0: libcamera-still -v 0 ...
Level 1: libcamera-still ... or libcamera-still -v 1
Level 2: libcamera-still -v ... or libcamera-still -v 2

### DIFF
--- a/core/options.hpp
+++ b/core/options.hpp
@@ -48,8 +48,8 @@ struct Options
 			 "Lists the available cameras attached to the system.")
 			("camera", value<unsigned int>(&camera)->default_value(0),
 			 "Chooses the camera to use. To list the available indexes, use the --list-cameras option.")
-			("verbose,v", value<bool>(&verbose)->default_value(false)->implicit_value(true),
-			 "Output extra debug and diagnostics")
+			("verbose,v", value<unsigned int>(&verbose)->default_value(1)->implicit_value(2),
+			 "Set verbosity level. Level 0 is no output, 1 is default, 2 is verbose.")
 			("config,c", value<std::string>(&config_file)->implicit_value("config.txt"),
 			 "Read the options from a file. If no filename is specified, default to config.txt. "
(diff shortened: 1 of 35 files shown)


core/options.hpp
```diff
@@ -48,8 +48,8 @@ struct Options
 			 "Lists the available cameras attached to the system.")
 			("camera", value<unsigned int>(&camera)->default_value(0),
 			 "Chooses the camera to use. To list the available indexes, use the --list-cameras option.")
-			("verbose,v", value<bool>(&verbose)->default_value(false)->implicit_value(true),
-			 "Output extra debug and diagnostics")
+			("verbose,v", value<unsigned int>(&verbose)->default_value(1)->implicit_value(2),
+			 "Set verbosity level. Level 0 is no output, 1 is default, 2 is verbose.")
 			("config,c", value<std::string>(&config_file)->implicit_value("config.txt"),
```
naushir, line 53: I wonder if we ought to disable libcamera level logging if we use ``-v 0``?

This can be done with
```
libcamera::logSetTarget(libcamera::LoggingTargetNone);
```
in ``LibcameraApp::OpenCamera()`` perhaps?

naushir:
I'm still in two minds about this... I wonder if we might want to define a macro such as:

```
#define LOG(lev, msg)				 \
do {								 \
	if (options_->verbose >= lev)	 \
		std::cerr << msg;	 		 \
} while(0)
```
and use that in place of ``if (options >= 2) ...``.  This seems neater, but we do run into the problem that ``options_->verbose`` is not available globally.  Maybe we can store the verbosity as a static member of ``LibcameraApp`` and access it globally with ``LibcameraApp::verbosity``?

PR #330, opened 2022-07-27
options: Use std::cout for --help, --list-cameras and --version output

naushir:

### DIFF
--- a/core/options.cpp
+++ b/core/options.cpp
@@ -71,14 +71,14 @@ bool Options::Parse(int argc, char *argv[])
 
 	if (help)
 	{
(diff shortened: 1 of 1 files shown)


naushir:
@davidplowman

davidplowman:
Yes, I guess that makes sense!

Are we happy to ignore the style checker here? I know we normally ignore it when dealing with options.cpp because it just doesn't understand the weird boost options syntax, but the complaints here are kind of "normal" ones. Happy to merge if you're OK with the formatting, though.

naushir:
Yes, I prefer the current format over the style checker suggestions here.

PR #335, opened 2022-08-04
Add const to input_fmt

jordemort:
Hey folks, ran into a compile error building this against ffmpeg 5.0 - the compiler really wanted the input_fmt to be const.

### DIFF
--- a/encoder/libav_encoder.cpp
+++ b/encoder/libav_encoder.cpp
@@ -129,7 +129,11 @@ void LibAvEncoder::initVideoCodec(VideoOptions const *options, StreamInfo const 
 
 void LibAvEncoder::initAudioInCodec(VideoOptions const *options, StreamInfo const &info)
 {
+#if LIBAVUTIL_VERSION_MAJOR < 58
 	AVInputFormat *input_fmt = (AVInputFormat *)av_find_input_format("pulse");
+#else
+	const AVInputFormat *input_fmt = (AVInputFormat *)av_find_input_format("pulse");
+#endif
 
 	assert(in_fmt_ctx_ == nullptr);
 	int ret = avformat_open_input(&in_fmt_ctx_, options->audio_device.c_str(), input_fmt, nullptr);
--- a/encoder/libav_encoder.hpp
+++ b/encoder/libav_encoder.hpp
@@ -22,6 +22,7 @@ extern "C"
 #include "libavutil/hwcontext.h"
 #include "libavutil/hwcontext_drm.h"
 #include "libavutil/timestamp.h"
+#include "libavutil/version.h"
 #include "libswresample/swresample.h"
 }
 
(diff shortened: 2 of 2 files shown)


naushir:
This is a bit of an annoying one.  Our current ffmpeg build v4.3 requires a non-const value to ``avformat_open_input``:
```
int avformat_open_input(AVFormatContext **ps, const char *url, ff_const59 AVInputFormat *fmt, AVDictionary **options);
```
so this change will break existing builds.  But as you noted, for ffmpeg v5.0 and later this will have to change to a const.

Unfortunately, I don't think we can take this change as-is yet as we are not ready to move to ffmpeg v5.0 in our releases just yet.

naushir:
You can see the compile error from the CI build [here](https://github.com/raspberrypi/libcamera-apps/runs/7687565735?check_suite_focus=true).

encoder/libav_encoder.cpp
```diff
@@ -129,7 +129,7 @@ void LibAvEncoder::initVideoCodec(VideoOptions const *options, StreamInfo const
 
 void LibAvEncoder::initAudioInCodec(VideoOptions const *options, StreamInfo const &info)
 {
-	AVInputFormat *input_fmt = (AVInputFormat *)av_find_input_format("pulse");
+	const AVInputFormat *input_fmt = (AVInputFormat *)av_find_input_format("pulse");
```
naushir, line 133: Perhaps things would work for all versions if you changed ``const`` to ``ff_const59``?

naushir: Probably not - see avformat.h:

```
    /**
     * The ff_const59 define is not part of the public API and will
     * be removed without further warning.
     */
```

naushir: You may be able to use the ``FF_API_AVIOFORMAT`` define, It's not clear what exactly that is meant to be used for, but might work for our problem here.

jordemort: It sounds like FF_API_AVIOFORMAT is also private, re: https://www.mail-archive.com/ffmpeg-devel@ffmpeg.org/msg81099.html

> Could you add a comment stating that, much like the FF_API_ defines, this ff_const59 define is not part of the public API and will eventually disappear without warning?

It seems like LIBAVUTIL_VERSION_MAJOR is intended for public consumption tho, so I've given that a shot.

naushir: Looks good.  Once the checks have passed, I'll merge this in.  Thanks!

naushir:
Actually, would you be able to re-submit this as a single patch instead of 2 commits?  Would be tidier in the history.

encoder/libav_encoder.cpp
jordemort: I'm not sure if it's correct but it passes your CI and also builds with 5.0 over here :)

jordemort:
> Actually, would you be able to re-submit this as a single patch instead of 2 commits? Would be tidier in the history.

Squashed down to one commit, I tried to write a better commit message for you too

naushir:
Great, I'll merge this as soon as the tests have completed.

PR #356, opened 2022-09-03
Watchdog for libcamera-apps.

tmsd2001:
The bash compares the CPU load and if the program hangs, it kills and restarts.


tmsd2001:
I just wanted to offer the watchdog,
I must have done something wrong there.

Issue #354, opened 2022-09-01
[question] Problem with OV5647


tmsd2001:
sometimes when using the libcamera-detect app I get a Segmentation fault.
With the imx477 I don't get any errors.

I think it might be because I'm using a longer than normal cable, but I can't test it because the camera is difficult to access.
The info-camera output:
```
Available cameras
-----------------
0 : ov5647 [2592x1944] (/base/soc/i2c0mux/i2c@1/ov5647@36)
    Modes: 'SGBRG10_CSI2P' : 640x480 [58.92 fps - (16, 0)/2560x1920 crop]
                             1296x972 [43.25 fps - (0, 0)/2592x1944 crop]
                             1920x1080 [30.62 fps - (348, 434)/1928x1080 crop]
                             2592x1944 [15.63 fps - (0, 0)/2592x1944 crop]
```

The OV5647 shows frequently this info, the lower the frame rate, the more common.
```
couch[62] (0.64) @ 331,349 357x330
TfStage: Inference time: 86655.5 ms
Stream configuration adjusted
[13:49:42.010022542] [1542966]  INFO Camera camera.cpp:1035 configuring streams: (0) 2592x1944-YUV420 (1) 2592x1944-SGRBG10_CSI2P
[13:49:42.011959138] [1542967]  INFO RPI raspberrypi.cpp:761 Sensor: /base/soc/i2c0mux/i2c@1/ov5647@36 - Selected sensor format: 2592x1944-SGRBG10_1X10 - Selected unicam format: 2592x1944-pgAA
TfStage: no low resolution stream
TfStage: Main stream is 2592x1944
 detected
[13:49:42.535174483] [1542966]  INFO Camera camera.cpp:1035 configuring streams: (0) 1296x972-YUV420 (1) 450x300-YUV420
[13:49:42.537029766] [1542967]  INFO RPI raspberrypi.cpp:761 Sensor: /base/soc/i2c0mux/i2c@1/ov5647@36 - Selected sensor format: 1296x972-SGRBG10_1X10 - Selected unicam format: 1296x972-pgAA
TfStage: Low resolution stream is 450x300
TfStage: Main stream is 1296x972
couch[62] (0.64) @ 331,349 357x327
```

What is your opinion, it is because of the cable and can I change it with software.
I need only a slow Framerate 0.5 - 5 fps
For example slow transfer rate

tmsd2001:
looks like the problem is with libcamera and not here.

[...]

PR #357, opened 2022-09-05
Watchdog for libcamera-apps.

tmsd2001:
try again, I hope this time it's right

### DIFF
--- a/utils/watchdog.sh
+++ b/utils/watchdog.sh
@@ -1,0 +1,36 @@
+#! /bin/bash
+
+datum=$(date +%Y%m%d_%H%M%S)
+PNAME="libcamera-detect" # Process to Watch
+load="$(/usr/bin/ps -C ${PNAME} -o %cpu | tail -1)"
+echo "step 1"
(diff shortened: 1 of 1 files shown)


davidplowman:
Hi, thanks for submitting this. It wasn't terribly clear to me what this script does, why I might need it and how I would use it. Could you explain some of that, please? Thanks.

tmsd2001:
In my case it often happens that the app no ​​longer runs, although the process is still active with 1% CPU use. In this case the Script kill the App and restart then. You can also use it to automatically start the app after boot , and other ways the camera stopped working.
thought for example for 24/7 cameras.
However, I already have another version.

davidplowman:
Thanks for the update. The first thing I'd like to do is understand the underlying problem better because it sounds like maybe something isn't working correctly. Are you able to give instructions that will provoke it and file an issue that we can then look at? Thanks!

tmsd2001:
I had already opened one.
[https://github.com/raspberrypi/libcamera-apps/issues/354#issue-1358593762](url)

PR #358, opened 2022-09-07
encoder: Add metadata output

will-v-pi:
Add the option to output metadata to either a file or stdout, similar to in libcamera-still.

Defaults to a json output for useability, but can be switched to a pure text output like libcamera-still using --metadata-plain.

Example usage:
`libcamera-raw -t 10000 --metadata metadata.json`
`libcamera-vid -t 10000 --metadata metadata.json`
`libcamera-vid -t 10000 --metadata metadata.txt --metadata-plain`

### DIFF
--- a/core/libcamera_encoder.hpp
+++ b/core/libcamera_encoder.hpp
@@ -45,6 +45,7 @@ public:
 			encode_buffer_queue_.push(completed_request); // creates a new reference
 		}
 		encoder_->EncodeBuffer(buffer->planes()[0].fd.get(), span.size(), mem, info, timestamp_ns / 1000);
+		encoder_->metadataReady(completed_request->metadata);
 	}
 	VideoOptions *GetOptions() const { return static_cast<VideoOptions *>(options_.get()); }
 	void StopEncoder() { encoder_.reset(); }
--- a/core/video_options.hpp
+++ b/core/video_options.hpp
@@ -58,6 +58,10 @@ struct VideoOptions : public Options
 			 "Write output to a circular buffer of the given size (in MB) which is saved on exit")
 			("frames", value<unsigned int>(&frames)->default_value(0),
 			 "Run for the exact number of frames specified. This will override any timeout set.")
+			("metadata", value<std::string>(&metadata),
+			 "Save captured image metadata to a file or \"-\" for stdout")
+			("metadata-plain", value<bool>(&metadata_plain)->default_value(false)->implicit_value(true),
+			 "Save the metadata as plaintext rather than json (requires --metadata)")
 #if LIBAV_PRESENT
 			("libav-format", value<std::string>(&libav_format)->default_value(""),
 			 "Sets the libav encoder output format to use. "
@@ -104,6 +108,8 @@ struct VideoOptions : public Options
 	uint32_t segment;
 	size_t circular;
 	uint32_t frames;
+	std::string metadata;
+	bool metadata_plain;
 
 	virtual bool Parse(int argc, char *argv[]) override
 	{
@@ -154,5 +160,6 @@ struct VideoOptions : public Options
 		std::cerr << "    split: " << split << std::endl;
 		std::cerr << "    segment: " << segment << std::endl;
 		std::cerr << "    circular: " << circular << std::endl;
+		std::cerr << "    metadata: " << metadata << std::endl;
 	}
 };
--- a/encoder/encoder.cpp
+++ b/encoder/encoder.cpp
@@ -16,6 +16,34 @@
 #include "libav_encoder.hpp"
 #endif
 
+Encoder::Encoder(VideoOptions const *options) : options_(options), buf_metadata_(std::cout.rdbuf()), of_metadata_()
+{
+	if (!options->metadata.empty())
+	{
+		const std::string &filename = options_->metadata;
+
+		if (filename.compare("-"))
+		{
+			of_metadata_.open(filename, std::ios::out);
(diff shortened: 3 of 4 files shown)


davidplowman:
Thanks for the quick turnaround!

One thing I'm wondering is whether we like the metadata being output when we start encoding a frame, but obviously the encoded frame can only be output when then encoder has finished, and the two things are normally asynchronous. I guess the only difference is that you might get an extra frame or two of metadata compared to the stream?

I suppose the alternative might be to push the metadata into the encode queue and output it when we give the encoded buffer to the output object. We could also make it the output object's job rather than the encoder's... maybe that feels nicer?

More trivially, I was also wondering about auto-selecting the metadata format based on the file extension? I know we don't tend to do that elsewhere, so it might seem a bit inconsistent, on the other hand it avoids any proliferation of `--metadata-plain` type options. Or maybe `--metadata-format plain` would be better as it's more extensible. Or perhaps `--metadata-format txt` because it matches the associated file type. We could auto-select based on the file type but also allow it to be overridden with `--metadata-format`. I think I'm tying myself in knots. What do folks think?

Finally, if we have fancy alternate metadata formats for videos, I'm thinking stills should have them too... what would that involve?

will-v-pi:
> One thing I'm wondering is whether we like the metadata being output when we start encoding a frame, but obviously the encoded frame can only be output when then encoder has finished, and the two things are normally asynchronous. I guess the only difference is that you might get an extra frame or two of metadata compared to the stream?

I haven't seen any occurrences of this in my testing, but yes, I suspect it could happen if the encoder drops a frame for some reason.

> I suppose the alternative might be to push the metadata into the encode queue and output it when we give the encoded buffer to the output object. We could also make it the output object's job rather than the encoder's... maybe that feels nicer?

I'll have a look at this, I wasn't sure how to move it to the output object as I couldn't see the best way to pass the metadata pointer around, but I could look at moving the writing of the metadata to the encodeBufferDone callback which occurs after each frame is encoded.

> More trivially, I was also wondering about auto-selecting the metadata format based on the file extension? I know we don't tend to do that elsewhere, so it might seem a bit inconsistent, on the other hand it avoids any proliferation of `--metadata-plain` type options. Or maybe `--metadata-format plain` would be better as it's more extensible. Or perhaps `--metadata-format txt` because it matches the associated file type. We could auto-select based on the file type but also allow it to be overridden with `--metadata-format`. I think I'm tying myself in knots. What do folks think?

I think --metadata-format would be the better option, with a default of json and the ability to set it to txt.

> Finally, if we have fancy alternate metadata formats for videos, I'm thinking stills should have them too... what would that involve?

Probably just making the function to save the metadata accessible to libcamera-still as well - I'll have a look.

will-v-pi:
I've implemented these changes with these commits (I can squash them once we're ready to merge), but I'll also take another look at moving the metadata output to the output class, as that would be the proper place for it.

core/libcamera_encoder.hpp
```diff
@@ -72,6 +72,8 @@ class LibcameraEncoder : public LibcameraApp
 			std::lock_guard<std::mutex> lock(encode_buffer_queue_mutex_);
 			if (encode_buffer_queue_.empty())
 				throw std::runtime_error("no buffer available to return");
+			auto completed_request = encode_buffer_queue_.front();
+			encoder_->metadataReady(completed_request->metadata);
```
davidplowman, line 76: I'm not sure about making it the Encoder's job to output metadata, it doesn't quite feel like a thing an Encoder should do. Can we make this a method of the output object?

core/video_options.hpp
```diff
@@ -58,6 +58,10 @@ struct VideoOptions : public Options
 			 "Write output to a circular buffer of the given size (in MB) which is saved on exit")
 			("frames", value<unsigned int>(&frames)->default_value(0),
 			 "Run for the exact number of frames specified. This will override any timeout set.")
+			("metadata", value<std::string>(&metadata),
+			 "Save captured image metadata to a file or \"-\" for stdout")
+			("metadata-plain", value<bool>(&metadata_plain)->default_value(false)->implicit_value(true),
+			 "Save the metadata as plaintext rather than json (requires --metadata)")
```
davidplowman: Was this going to become "metadata-format"? I think the "still" version has that, so probably make this one the same. Maybe this option just vanishes from here in favour of one in the base class.

will-v-pi: Yes, sorry, that's fixed in the second commit

apps/libcamera_raw.cpp
```diff
@@ -28,8 +28,9 @@ class LibcameraRaw : public LibcameraEncoder
 static void event_loop(LibcameraRaw &app)
 {
 	VideoOptions const *options = app.GetOptions();
-	std::unique_ptr<Output> output = std::unique_ptr<Output>(Output::Create(options));
+	std::shared_ptr<Output> output = std::shared_ptr<Output>(Output::Create(options));
```
davidplowman, line 31: Just wondering if this needs to be a shared_ptr? We're passing a raw pointer to MetadataReady so I wouldn't have thought anyone is grabbing an extra reference?

davidplowman:
Thanks, this is looking good to me now! Just a couple of little memory related questions to have a quick think about.

apps/libcamera_raw.cpp
will-v-pi: Yes, you're right, I'll rectify that

davidplowman:
I'm going to close this one because I'm going to tidy up one or two things and then I'll re-push.

PR #362, opened 2022-09-14
William's metadata output for video

davidplowman:
These are William's commits that add metadata output for video streams, plus a couple more commits to add the same metadata formatting options to stills capture, and also to add some tests.

Things ended up a bit awkward because we wanted to get the Output objects to write the metadata, rather than hold up any other threads, but things are sufficiently compartmentalised that getting the metadata down there was slightly annoying. We've gone with copying the metadata into a queue in the output object, though an alternative would have been to hold on to a reference to the CompletedRequest instead.

On balance I think copying the metadata is OK - it only does it when it really has been requested, and holding onto the CompletedRequest might delay recycling the request back to libcamera which might also be undesirable.

### DIFF
--- a/apps/CMakeLists.txt
+++ b/apps/CMakeLists.txt
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 3.6)
 
 project(libcamera-still)
 add_executable(libcamera-still libcamera_still.cpp)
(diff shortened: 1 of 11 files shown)


davidplowman:
@naushir Maybe one for you to have a look at and see what you think?

PR #366, opened 2022-09-22
core: Fix a leaking request when preview windows are closed

davidplowman:
Starting and stopping the preview window actually causes a camera request object to leak, and will ultimately lead to libcamera stalling. Not a use case we've ever considered before, but we may as well make it work!

### DIFF
--- a/core/libcamera_app.cpp
+++ b/core/libcamera_app.cpp
@@ -777,6 +777,7 @@ void LibcameraApp::stopPreview()
 	}
 	preview_thread_.join();
 	preview_item_ = PreviewItem();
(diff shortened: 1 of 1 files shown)


naushir:
Do you think this might be the cause of https://github.com/raspberrypi/libcamera-apps/issues/361?  Although we are not running with preview in that test.

davidplowman:
I don't think so. It's true that even though there's no preview window we do still run the preview thread. But firstly, the NullPreview drops the CompletedRequests immediately in its Show method so they can't hang around, and secondly doing a camera stop/start recreates all the requests anyway and therefore hides the problem. So alas, no.

Issue #368, opened 2022-09-27
Build fails with clang++ 15.0.1


tewarid:
Build fails at the following line of code when using clang++ 15.0.1 due to error `fatal error: variable 'count' set but not used [-Wunused-but-set-variable]`

https://github.com/raspberrypi/libcamera-apps/blob/dd97618a25523c2c4aa58f87af5f23e49aa6069c/apps/libcamera_jpeg.cpp#L42

naushir:
I'm a bit surprised at this warning, given count is incremented in the same statement... but the compiler is probably correct given it is never referenced in the loop.

Feel free to change this to ``for (;;)`` to fix this.

davidplowman:
I've merged the patch so will close this issue. Thanks!

PR #369, opened 2022-09-27
Fixes build failure with clang++ 15.0.1 #368

tewarid:


davidplowman:
Thanks for sending the fix!

PR #380, opened 2022-10-04
libav: Correctly pipe to stdout

naushir:
Replace the stdout filename "-" with "pipe:" when using libav.


naushir:
@davidplowman a small change to look at.

PR #403, opened 2022-11-08
Mode selection for framerate

naushir:
- options: Switch to std::optional for framerate
- core: Use automatic mode selection for framerate

### DIFF
--- a/core/libcamera_app.cpp
+++ b/core/libcamera_app.cpp
@@ -410,9 +410,9 @@ void LibcameraApp::StartCamera()
 		if (StillStream())
 			controls_.set(controls::FrameDurationLimits,
 						  libcamera::Span<const int64_t, 2>({ INT64_C(100), INT64_C(1000000000) }));
-		else if (options_->framerate > 0)
+		else if (!options_->framerate || options_->framerate.value() > 0)
 		{
-			int64_t frame_time = 1000000 / options_->framerate; // in us
+			int64_t frame_time = 1000000 / options_->framerate.value_or(DEFAULT_FRAMERATE); // in us
 			controls_.set(controls::FrameDurationLimits,
 						  libcamera::Span<const int64_t, 2>({ frame_time, frame_time }));
 		}
--- a/core/libcamera_app.cpp
+++ b/core/libcamera_app.cpp
@@ -11,6 +11,7 @@
 #include "core/libcamera_app.hpp"
 #include "core/options.hpp"
 
+#include <cmath>
 #include <fcntl.h>
 
 #include <sys/ioctl.h>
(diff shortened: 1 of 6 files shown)


core/libcamera_app.cpp
```diff
@@ -410,9 +410,9 @@ void LibcameraApp::StartCamera()
 		if (StillStream())
 			controls_.set(controls::FrameDurationLimits,
 						  libcamera::Span<const int64_t, 2>({ INT64_C(100), INT64_C(1000000000) }));
-		else if (options_->framerate > 0)
+		else if (!options_->framerate || options_->framerate.value() > 0)
 		{
-			int64_t frame_time = 1000000 / options_->framerate; // in us
+			int64_t frame_time = 1000000 / options_->framerate.value_or(30); // in us
```
davidplowman, line 507: I guess the only slight feeling I get reading this is that there's quite a lot of `value_or(30)` creeping in around the place.  You might think, why  30? Could I use something else? And so on.

Maybe the `Options` class should have a `get_framerate()` method which does that? And the `framerate` and `framerate_` might be private? Or maybe that is just classic C++ over-engineering! I'm not sure... and in this case not all that bothered either, I'm fine with it as it is too.

core/libcamera_app.cpp
```diff
@@ -126,6 +127,32 @@ void LibcameraApp::OpenCamera()
...
+				config->at(0).pixelFormat = pix;
+				config->validate();
+				camera_->configure(config.get());
+				auto fd_ctrl = camera_->controls().find(&controls::FrameDurationLimits);
+				sensor_modes_.emplace_back(size, pix, 1.0e6 / fd_ctrl->second.min().get<int64_t>());
+			}
+		}
+
+		libcamera::logSetLevel("RPI", "INFO");
+		libcamera::logSetLevel("Camera", "INFO");
```
davidplowman, line 154: It strikes me that this is pretty similar to what `--list-cameras` does, but I expect there's no very convenient way to share the code? Just thinking out loud.

core/libcamera_app.cpp
naushir: Yes, this bothers me a bit too!

core/libcamera_app.cpp
```diff
@@ -144,13 +171,71 @@ void LibcameraApp::CloseCamera()
...
+		// Penalise non-exact matches.
+		if (actual != desired)
+			score *= 2;
+
+		return score;
+	};
+
+	constexpr float penalty_AR = 1500.0;
+	constexpr float penalty_BD = 500.0;
+	constexpr float penalty_FPS = 2000.0;
```
davidplowman, line 191: Hehe. I shall wait for the feature request to expose these on the command line! :)

core/libcamera_app.cpp
naushir: Slightly different to the ``--list-cameras`` case.  In that we set the log target to none, so no messages will ever get printed.  Here we still get some error messages if something goes wrong in libcamera.  Maybe I should change the former?

core/libcamera_app.cpp
```diff
@@ -144,13 +171,71 @@ void LibcameraApp::CloseCamera()
...
+		double reqAr = static_cast<double>(req.width) / req.height;
+		double fmtAr = static_cast<double>(mode.size.width) / mode.size.height;
+
+		// This is a really ugly way of getting the bit depth of the format.
+		// But apart from duplicating the massive bayer format table, there is
+		// no other way to determine this.
+		std::string fmt = mode.format.toString();
+		unsigned int mode_depth = fmt.find("8") != std::string::npos ? 8 :
+								  fmt.find("10") != std::string::npos ? 10 :
+								  fmt.find("12") != std::string::npos ? 12 : 16;
```
davidplowman, line 208: Would it be tidier to hide this in a `depth()` method of the `SensorMode`? Again, don't really mind...

core/libcamera_app.cpp
davidplowman: If it's not straightforward, I would say to leave it...

core/libcamera_app.cpp
naushir: Yes, I can do that!

core/libcamera_app.cpp
```diff
@@ -205,6 +290,9 @@ void LibcameraApp::ConfigureViewfinder()
 		configuration_->at(lores_stream_num).bufferCount = configuration_->at(0).bufferCount;
 	}
 
+	if (select_mode)
+		options_->viewfinder_mode = selectModeForFramerate(size, options_->framerate.value());
+
```
davidplowman, line 295: I wonder if the `select_mode` flag could disappear if we did this at the top of the function (where we set `select_mode`)? Only a teeny thing, though. Same might apply elsewhere.

core/libcamera_app.hpp
```diff
@@ -184,6 +200,7 @@ class LibcameraApp
 	void stopPreview();
 	void previewThread();
 	void configureDenoise(const std::string &denoise_mode);
+	Mode selectModeForFramerate(const libcamera::Size size, double fps);
```
davidplowman, line 203: I guess `&size` is actually no better than `size` for a small structure like this, is it?

core/libcamera_app.cpp
naushir: Can't do this unfortunately, we need the "default" size to be available, and that only happens later on in the function.

core/libcamera_app.hpp
naushir: Shouldn't matter too much, but I'll switch to using a reference.

core/libcamera_app.cpp
naushir: I'll leave it for now.

PR #410, opened 2022-11-15
Add YUV save to stdout on "-" path

ivolucas:
The other save modes have the option of setting a '-' path to redirect to stdout the YUV does not.
This normalizes the behavior


naushir:
Thank you for this fix!  Merging now.

PR #411, opened 2022-11-17
Add libcamera autofocus options

ivolucas:
Libcamera has some controls options for focus that are useful for some new cameras that exists. 
Arducam 16Mp and 64Mp cameras with autofocus are examples of this. 

This commit add some of this options for focus to the command line.

### DIFF
--- a/core/libcamera_app.cpp
+++ b/core/libcamera_app.cpp
@@ -381,10 +381,9 @@ void LibcameraApp::ConfigureVideo(unsigned int flags)
 	// Now we get to override any of the default settings from the options_->
 	StreamConfiguration &cfg = configuration_->at(0);
 	cfg.pixelFormat = libcamera::formats::YUV420;
-	if (options_->buffer_count > 0)
-	{
+	cfg.bufferCount = 6; // 6 buffers is better than 4
+	if (options_->buffer_count > 0) // give option to the user to use a different value
 		cfg.bufferCount = options_->buffer_count;
-	}
 	if (options_->width)
 		cfg.size.width = options_->width;
 	if (options_->height)
@@ -504,8 +503,6 @@ void LibcameraApp::StartCamera()
 		controls_.set(controls::AfMetering, controls::AfMeteringWindows);
 		//set window
 		controls_.set(controls::AfWindows, afwindows_rectangle);
-
-		
 	}
 
 	// Framerate is a bit weird. If it was set programmatically, we go with that, but
@@ -548,15 +545,20 @@ void LibcameraApp::StartCamera()
 	if (!controls_.get(controls::Sharpness))
 		controls_.set(controls::Sharpness, options_->sharpness);
 
-	if (!controls_.get(controls::AfMode))
-		controls_.set(controls::AfMode, options_->afMode_index);
-	if (!controls_.get(controls::AfRange))
-		controls_.set(controls::AfRange, options_->afRange_index);
-	if (!controls_.get(controls::AfSpeed))
-		controls_.set(controls::AfSpeed, options_->afSpeed_index);
-	if (!controls_.get(controls::LensPosition))
-		controls_.set(controls::LensPosition, options_->lens_position);
-
+	if(camera_->controls().count(&controls::AfMode)>0){
+		LOG(2, "Camera has AfMode");
+		if (options_->afMode_index!=-1 && !controls_.get(controls::AfMode)){
+			controls_.set(controls::AfMode, options_->afMode_index);
+			if(options_->afMode_index == controls::AfModeAuto){
+				//does the first focus on start 
+				controls_.set(controls::AfTrigger, controls::AfTriggerStart);
+			}
+		}
+		if (options_->afRange_index!=-1 && !controls_.get(controls::AfRange))
+			controls_.set(controls::AfRange, options_->afRange_index);
+		if (options_->afSpeed_index!=-1 && !controls_.get(controls::AfSpeed))
+			controls_.set(controls::AfSpeed, options_->afSpeed_index);
+	}
 	if (camera_->start(&controls_))
 		throw std::runtime_error("failed to start camera");
 	controls_.clear();
--- a/core/libcamera_app.cpp
+++ b/core/libcamera_app.cpp
@@ -381,7 +381,10 @@ void LibcameraApp::ConfigureVideo(unsigned int flags)
 	// Now we get to override any of the default settings from the options_->
 	StreamConfiguration &cfg = configuration_->at(0);
 	cfg.pixelFormat = libcamera::formats::YUV420;
-	cfg.bufferCount = 6; // 6 buffers is better than 4
+	if (options_->buffer_count > 0)
+	{
+		cfg.bufferCount = options_->buffer_count;
+	}
 	if (options_->width)
 		cfg.size.width = options_->width;
 	if (options_->height)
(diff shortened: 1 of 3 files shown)


naushir:
Thank you for this work.  Most of it looks reasonable, but I'd like to address a few minor details in the code.

core/libcamera_app.cpp
```diff
@@ -381,7 +381,10 @@ void LibcameraApp::ConfigureVideo(unsigned int flags)
 	// Now we get to override any of the default settings from the options_->
 	StreamConfiguration &cfg = configuration_->at(0);
 	cfg.pixelFormat = libcamera::formats::YUV420;
-	cfg.bufferCount = 6; // 6 buffers is better than 4
+	if (options_->buffer_count > 0)
+	{
+		cfg.bufferCount = options_->buffer_count;
+	}
```
naushir, line 387: The reasoning behind this change is reasonable.  However, we are working to integrate something like this in the core of libcamera.  You can see the patches [here](https://patchwork.libcamera.org/project/libcamera/list/?series=3555).  I would prefer if we did not add similar code to libcamera-apps.

core/libcamera_app.cpp
```diff
@@ -485,6 +488,26 @@ void LibcameraApp::StartCamera()
...
+		afwindows_rectangle[0]= Rectangle(x, y, w, h);
+		afwindows_rectangle[0].translateBy(sensor_area.topLeft());
+		LOG(2, "Using AfWindow " << afwindows_rectangle[0].toString());
+		//activate the AfMeteringWindows
+		controls_.set(controls::AfMetering, controls::AfMeteringWindows);
+		//set window
+		controls_.set(controls::AfWindows, afwindows_rectangle);
+
+		
+	}
```
naushir, line 509: Could you trim the empty spaces please?

core/libcamera_app.cpp
```diff
@@ -525,6 +548,15 @@ void LibcameraApp::StartCamera()
...
 
+	if (!controls_.get(controls::AfMode))
+		controls_.set(controls::AfMode, options_->afMode_index);
+	if (!controls_.get(controls::AfRange))
+		controls_.set(controls::AfRange, options_->afRange_index);
+	if (!controls_.get(controls::AfSpeed))
+		controls_.set(controls::AfSpeed, options_->afSpeed_index);
+	if (!controls_.get(controls::LensPosition))
+		controls_.set(controls::LensPosition, options_->lens_position);
+
```
naushir, line 559: On thing we ought to probably do (here and above where we call controls_.set()) is to ensure we only set AF related controls when there is an actual lens present.

One possible way to test this is to do ``camera_->controls().find(&controls::AfMode)``.  This relies on Arudcam doing the right thing in their closed source modules and only registering controls::AfMode if there is a lens present.

core/options.cpp
```diff
@@ -298,6 +325,14 @@ void Options::Print() const
 	std::cerr << "    lores-width: " << lores_width << std::endl;
 	std::cerr << "    lores-height: " << lores_height << std::endl;
 
+	std::cerr << "    autofocus-mode: " << afMode << std::endl;
```
naushir, line 328: Again, this should only be printed if there is a controllable lens present.

core/options.hpp
```diff
@@ -133,6 +133,17 @@ struct Options
...
+			("autofocus-mode", value<std::string>(&afMode)->default_value("manual"),
+			 " Control to set the mode of the AF (autofocus) algorithm.(manual, auto, continuous)")
+			("autofocus-range", value<std::string>(&afRange)->default_value("normal"),
+			 "Set the range of focus distances that is scanned.(normal, macro, full)")
+			("autofocus-speed", value<std::string>(&afSpeed)->default_value("normal"),
+			 "Control that determines whether the AF algorithm is to move the lens as quickly as possible or more steadily.(normal, fast)")
+			("autofocus-window", value<std::string>(&afWindow)->default_value("0,0,0,0"), 
+			"Sets AfMetering to  AfMeteringWindows an set region used for examle  e.g. 0.25,0.25,0.5,0.5")
+			("lens-position", value<float>(&lens_position)->default_value(0),
+			 "Acts as a control to instruct the lens to move to a particular position and also reports back the position of the lens for each frame. (0 moves the lens to infinity;0.5 moves the lens to twice the hyperfocal distance;1 moves the lens to the hyperfocal position )")
```
naushir, line 146: We are in discussion about the exact units for the lens-position control.  Right now it is based on hyperfocal distance, but we probably will recommend changing to diopters.  I suggest you remove this for now to avoid any confusion.

core/options.hpp
```diff
@@ -193,8 +205,18 @@ struct Options
...
+	unsigned int buffer_count;
+	std::string afRange;
+	int afRange_index;
+	std::string afSpeed;
+	int afSpeed_index;
+	std::string afWindow;
+	float afWindow_x, afWindow_y, afWindow_width, afWindow_height;
 	std::string metadata;
 	std::string metadata_format;
+
```
naushir, line 219: You can remove this blank line.

naushir:
@njhollinghurst might have a few further comments.

core/libcamera_app.cpp
njhollinghurst: `AfModeAuto` has (or should have) no effect without `AfTrigger`. It seems reasonable to trigger it when the app starts, but in that case, those controls should be in OpenCamera() instead of StartCamera() to avoid repeated triggering on mode change.

I don't think we should set `LensPosition` by default, unless this control exists _and_ the user explicitly requests that it be set to an explicit value.

davidplowman: I'd quite like us to define the behaviours of the various libcamera-apps in the presence of autofocus controls, and be clear what happens in the LibcameraApp methods, and what the applications must implement.

For example, libcamera-hello/libcamera-vid are pretty straightforward. They set the AfMode to the one requested, the lens position if one is explicitly given. If an AfTrigger is requested, that could do a single sweep on startup. We could do all this in StartCamera, along with all the other controls.

libcamera-still is trickier. For example, we could say that when you switch into stills mode, it completely ignores all the AF controls.  Then it's up to the application code to do things like run an AF cycle before switching to capture. If it's in CAF mode, possibly it has to pause before switching (or is that unnecessary)? Would that cover the most obvious use-cases? But this is all open to debate, you get the idea!

There will certainly be scenarios  that are not covered by a set of simple options like this. Someone will come along and say "I only want it to do an AF sweep on every 10th timelapse capture", or something (OK, that's a bit extreme, but you know what I mean!). I'd prefer not to over-complicate the libcamera-apps, and I certainly don't want to see interactive features (like "press a key to turn AF on"). There's already a solution for all that (Picamera2).

What do folks think?

core/options.hpp
naushir: For now I would remove the option until the actual definition is finalised by libcamera.

naushir: To clarify, this is an option that should be in once the definition has been finalised by libcamera.

core/libcamera_app.cpp
njhollinghurst: Most of these AF controls are "set and forget". I think it does makes sense to put them in either StartCamera() or OpenCamera(), for all applications to use.

The awkward ones are `AfPause`, alluded to above, and `AfTrigger`. Contrary to my previous comment, perhaps both of those should be for the application to control. We might also want to add an option for libcamera-still to wait until AF is stable (state _not_ equal to AfStateScanning) before capturing a still; but those things could be added in a later commit.

core/libcamera_app.cpp
naushir: Given that this change is not related to AF, would you be able to create a new PR for this?  We can then merge it straight away.

davidplowman: I think we should probably add this change to all the `ConfigureXxx` methods? That way it's a consistent option across the board.

davidplowman: Though having said that... maybe the viewfinder mode wants a separate `--viewfinder-buffer-count` option, because you might want a different number from the still option. In fact you certainly would.

core/libcamera_app.cpp
```diff
@@ -548,15 +545,20 @@ void LibcameraApp::StartCamera()
...
+			if(options_->afMode_index == controls::AfModeAuto){
+				//does the first focus on start 
+				controls_.set(controls::AfTrigger, controls::AfTriggerStart);
+			}
+		}
+		if (options_->afRange_index!=-1 && !controls_.get(controls::AfRange))
+			controls_.set(controls::AfRange, options_->afRange_index);
+		if (options_->afSpeed_index!=-1 && !controls_.get(controls::AfSpeed))
+			controls_.set(controls::AfSpeed, options_->afSpeed_index);
+	}
```
njhollinghurst, line 561: Thanks for adding `AfTrigger`, but I think it should not be in `StartCamera()`. The problem is that it will re-trigger when capturing a still, without waiting for completion. Depending on the implementation, it might result in blurry image.

We could make it conditional on first start or on stream type... but rather than trying to 'guess' the application's intention, it would be cleaner to deal with it at the individual application level.

Could you please remove `AfTrigger` for now. It will be re-worked in a later commit.

core/libcamera_app.cpp
ivolucas: I will remove this from this PR and add the suggested --viewfinder-buffer-count

core/libcamera_app.cpp
ivolucas: done

core/libcamera_app.cpp
ivolucas: done

naushir:
This one is almost ready to merge. 
The style checker has picked out some issues that need correcting:
https://github.com/raspberrypi/libcamera-apps/actions/runs/3514039845/jobs/5906783036

You can ignore the warnings about the std::map definitions, and most of the others are likely asking for spaces to tabs conversions.

naushir:
Thanks for the fixups!
We don't use merge commits on our tree, but  I'll fix this locally and push your changes shortly.

PR #412, opened 2022-11-21
Option to define buffer count overwrite

ivolucas:
Currently libcamera-apps overwrites the bufferCount configuration supplied by the libcamera implementation with a fix value of six.
This alteration removes the harcode overwrites of the value by configuration options on the command line.

To keep current behavior we can change the default configuration from 0 to 6 (the previous hardcoded value) but I think this value should not be overwritten by default, the value return by the implementation of the libcamerra for the current camera device should be always the default in my opinion.

### DIFF
--- a/core/libcamera_app.cpp
+++ b/core/libcamera_app.cpp
@@ -381,7 +381,8 @@ void LibcameraApp::ConfigureVideo(unsigned int flags)
 	// Now we get to override any of the default settings from the options_->
 	StreamConfiguration &cfg = configuration_->at(0);
 	cfg.pixelFormat = libcamera::formats::YUV420;
-	cfg.bufferCount = 6; // 6 buffers is better than 4
+	if (options_->buffer_count > 0)
+		cfg.bufferCount = options_->buffer_count;
 	if (options_->width)
 		cfg.size.width = options_->width;
 	if (options_->height)
@@ -419,7 +420,9 @@ void LibcameraApp::ConfigureVideo(unsigned int flags)
 			throw std::runtime_error("Low res image larger than video");
 		configuration_->at(lores_index).pixelFormat = libcamera::formats::YUV420;
 		configuration_->at(lores_index).size = lores_size;
-		configuration_->at(lores_index).bufferCount = configuration_->at(0).bufferCount;
+		if(options_->viewfinder_buffer_count > 0)
+			configuration_->at(lores_index).bufferCount = options_->viewfinder_buffer_count;
+		
 	}
 	configuration_->transform = options_->transform;
 
--- a/core/options.hpp
+++ b/core/options.hpp
@@ -133,6 +133,8 @@ struct Options
 			 "Camera mode as W:H:bit-depth:packing, where packing is P (packed) or U (unpacked)")
 			("viewfinder-mode", value<std::string>(&viewfinder_mode_string),
 			 "Camera mode for preview as W:H:bit-depth:packing, where packing is P (packed) or U (unpacked)")
+			("buffer-count", value<unsigned int>(&buffer_count)->default_value(0), "overwrite bufferCount config from libcamera for video; raw and still (0 to use default provided by the libcamera implementation)")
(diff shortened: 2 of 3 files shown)


core/libcamera_app.cpp
```diff
@@ -381,7 +381,8 @@ void LibcameraApp::ConfigureVideo(unsigned int flags)
 	// Now we get to override any of the default settings from the options_->
 	StreamConfiguration &cfg = configuration_->at(0);
 	cfg.pixelFormat = libcamera::formats::YUV420;
-	cfg.bufferCount = 6; // 6 buffers is better than 4
+	if (options_->buffer_count > 0)
+		cfg.bufferCount = options_->buffer_count;
```
naushir, line 385: I think this logic might be a bit wrong.  Do we instead want to do something like:

```
cfg.bufferCount = 6; // 6 buffers is better than 4
if (options_->buffer_count > 0)
    cfg.bufferCount = options_->buffer_count;
```

Ditto for the change below.

core/options.hpp
```diff
@@ -133,6 +133,8 @@ struct Options
 			 "Camera mode as W:H:bit-depth:packing, where packing is P (packed) or U (unpacked)")
 			("viewfinder-mode", value<std::string>(&viewfinder_mode_string),
 			 "Camera mode for preview as W:H:bit-depth:packing, where packing is P (packed) or U (unpacked)")
+			("buffer-count", value<unsigned int>(&buffer_count)->default_value(0), "overwrite bufferCount config from libcamera for video; raw and still (0 to use default provided by the libcamera implementation)")
+			("viewfinder-buffer-count", value<unsigned int>(&viewfinder_buffer_count)->default_value(0), "overwrite bufferCount config from libcamera  for the preview (0 to use default provided by the libcamera implementation)")
```
naushir, line 137: Could you re-phrase the description a bit.  Something like
```
Number of in-flight requests (and buffers) configured for video; raw; and still.
```

core/libcamera_app.cpp
naushir: It's better to explicitly set the value 6 as before.  I'm not sure what the pipeline handler defaults to, but if that number changes in the future, it should not affect libcamera-apps if we keep this value at a default of 6.

ivolucas: For what it defaults to, chek this for raspberry pi cameras. https://github.com/raspberrypi/libcamera/blob/7c855784a28d0c9b742c6137bf311ac3779373df/src/libcamera/pipeline/raspberrypi/raspberrypi.cpp#L600

What if I change the default of the options to 6 inset of 0? t
This way we keep the current behavior, and give the option of the user to use the original defaults

naushir: > What if I change the default of the options to 6 inset of 0? t

That would be fine with me.

core/libcamera_app.cpp
```diff
@@ -419,7 +420,9 @@ void LibcameraApp::ConfigureVideo(unsigned int flags)
 			throw std::runtime_error("Low res image larger than video");
 		configuration_->at(lores_index).pixelFormat = libcamera::formats::YUV420;
 		configuration_->at(lores_index).size = lores_size;
-		configuration_->at(lores_index).bufferCount = configuration_->at(0).bufferCount;
+		if(options_->viewfinder_buffer_count > 0)
+			configuration_->at(lores_index).bufferCount = options_->viewfinder_buffer_count;
+
```
naushir, line 425: This is for the low res image - different from viewfinder.  I think you need to set the viewfinder buffer count a few lines above - ``configuration_->at(0).bufferCount = ...``.

core/options.hpp
naushir: Looking at the changes again, the RAW stream buffer count is not updated from the command line argument.

naushir:
This looks ready to go with some minor changes.  You will also get some warnings from the style checker here: https://github.com/raspberrypi/libcamera-apps/actions/runs/3523570993/jobs/5907958442

Once everything is ready, I'll hit the merge button.

naushir:
Thanks for the fixes.  The style checker is still showing some issues I'm afraid:
https://github.com/raspberrypi/libcamera-apps/actions/runs/3525215005/jobs/5922553571

I'll fix them up manually and merge this shortly.

PR #415, opened 2022-11-29
core: LibcameraApp::SetControls has a "merge" option, true by default

njhollinghurst:
This allows new controls to be added to the stored list rather than replacing it. As before, the argument is invalidated.

### DIFF
--- a/core/libcamera_app.cpp
+++ b/core/libcamera_app.cpp
@@ -730,10 +730,15 @@ void LibcameraApp::ShowPreview(CompletedRequestPtr &completed_request, Stream *s
 	preview_cond_var_.notify_one();
 }
 
-void LibcameraApp::SetControls(ControlList &controls)
+void LibcameraApp::SetControls(const ControlList &controls)
 {
 	std::lock_guard<std::mutex> lock(control_mutex_);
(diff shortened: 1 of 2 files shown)


njhollinghurst:
This is something we discussed, though I'm not sure if we really need it. Currently nobody calls ``LibcameraApp::SetControls()``

davidplowman:
Yep, LGTM. I guess I'm mildly amused that no one even calls this thing yet, but presumably we will start doing so soon with some of our other updates? Let me know when you'd like me to push the Big Green Button!

njhollinghurst:
Yes, I am calling it now and it seems to work, but I guess a bit more testing (e.g. to check that the merge precedence works the correct way around) would not go amiss. I'll do that today.

njhollinghurst:
Yikes. A precondition of ``ControlList::merge()`` is that
```
* Only control lists created from the same ControlIdMap or ControlInfoMap may
 * be merged. Attempting to do otherwise results in undefined behaviour.
```
My change would require an application-writer to always create temporary control lists with the same ``ControlIdMap`` value that ``LibcameraApp`` used for its internal list, or risk a crash. This is obviously not convenient for anybody.

PR #417, opened 2022-11-30
More autofocus controls

njhollinghurst:
Added the ``--lens-position`` option, based on PR #411 but made optional and with the additional option to use the camera's default value. (The help text is deliberately vague about the units of measurement.)

Added AfTrigger, inferred when AfMode is Auto, but not for stills mode (to avoid moving the lens during capture).

The second commit modifies ``libcamera-still`` to trigger AF and wait for it to complete before capturing each image; the initial viewfinder session is forced to AfModeManual and therefore supports a LensPosition setting.

Frankly, this all seems a bit overwrought and I do wonder if it could be simpler?

### DIFF
--- a/core/options.hpp
+++ b/core/options.hpp
@@ -39,7 +39,7 @@ struct Mode
 
 struct Options
 {
-	Options() : options_("Valid options are", 120, 80)
+	Options() : set_default_lens_position(false), options_("Valid options are", 120, 80)
 	{
 		using namespace boost::program_options;
 		// clang-format off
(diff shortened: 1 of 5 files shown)


core/options.hpp
```diff
@@ -144,6 +144,8 @@ struct Options
 			 "Control that determines whether the AF algorithm is to move the lens as quickly as possible or more steadily.(normal, fast)")
 			("autofocus-window", value<std::string>(&afWindow)->default_value("0,0,0,0"),
 			"Sets AfMetering to  AfMeteringWindows an set region used, e.g. 0.25,0.25,0.5,0.5")
+			("lens-position", value<std::string>(&lens_position_)->default_value(""),
+			 "Set the lens to a particular focus position, expressed as a reciprocal distance (0 moves the lens to infinity), or \"default\" for the hyperfocal distance")
```
naushir, line 148: Should we perhaps use "hyperfocal" explicitly here instead of default? I'm not really fussed either way.

njhollinghurst: Hyperfocal: I didn't want to pre-empt the libcamera specification so I was keeping it a bit vague, but could change this.

naushir: Good point, maybe best leave as "default" for now.

njhollinghurst:
I think the above version was a bit premature. I'm going to remove some of the auto-inferred magic from ``libcamera-still``, and have it do what it's told during viewfinder; an explicit option ``--autofocus-on-capture`` will then enable the switch to auto mode just before capture.

njhollinghurst:
I've tidied up ``libcamera-still`` some more, and added a timeout just in case the IPA fails to pass through the Scanning state for every invocation of AfTriggerStart (e.g. if it thinks it's already well-focused or doomed to failure). I think this is good to merge now.

PR #420, opened 2022-12-01
output: null_encoder: Fix bug in ordering of input/output callbacks

naushir:
This would cause a pop from the metadata queue in the output callback before the input callback has a chance to push to the queue.


naushir:
@davidplowman

davidplowman:
The only thing I might do is put a little comment in the code just to warn anyone who feels tempted to mess around at a later date? Because this is a slightly nasty trap otherwise. Apart from that, LGTM, just push the merge button when you're happy!

PR #425, opened 2022-12-23
libcamera-detect: add multiple file name output generation options

ThingEngineer:
Added file name generation similar to libcamera-still including timestamp and datetime. Also added a the --timeformat argument to allow passing in custom strftime() date/time format strings when the --datetime option is true. The default value for timeformat is: "%m%d%H%M%S". The file extension is set by options->encoding for the datetime and timestamp options. 
The default --output option retain its previous functionality. 

Examples:
 -o /home/user/capture/car- --datetime
     output file: car-1223033452.jpg

 -o /home/user/capture/car- --timestamp --timeformat "%F %r"
     output file: car-2022-12-23 03:34:52 PM.jpg

 -o /home/user/capture/ --timestamp
     output file: 1671831417.jpg

-o /home/user/capture/car%03d.jpg (Default/Existing)
     output file: car001.jpg

### DIFF
--- a/.github/workflows/libcamera-apps-test.yml
+++ b/.github/workflows/libcamera-apps-test.yml
@@ -82,7 +82,7 @@ jobs:
 
     strategy:
       matrix:
-        camera: [ imx219, imx477 ]
+        camera: [ imx219, imx477, imx708 ]
 
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


naushir:
@ThingEngineer, thanks for the patch!
Functionally, it looks fine and I am happy to merge it.  Before I do, could you squash both commits down into one as they are related, and could you re-format the commit message please.  We need a small title with an appropriate prefix (probably "libcamera_detect:" in this case), and a small description of the change.  Have a look at the git log to see examples of other commit message formats.

PR #431, opened 2023-01-03
core: Tweak default AF options; AfMode now defaults typically to CAF

njhollinghurst:
AF Mode now defaults to the highest supported mode (typically CAF) unless an explicit lens position or af_on_capture was requested. Speed and Range now default to "normal" rather than unset.


naushir:
LGTM

PR #432, opened 2023-01-03
core: Add HDR option, with a stop-gap implementation using V4L2 control

njhollinghurst:


core/options.hpp
```diff
@@ -220,6 +222,7 @@ struct Options
 	bool set_default_lens_position;
 	std::string metadata;
 	std::string metadata_format;
+	int hdr;
```
naushir, line 226: I wonder if we should make hdr a bool, then define the option like
```
("hdr", value<bool>(&hdr)->default_value(false)->implicit_value(true),
"Enable (1) or disable (0) High Dynamic Range, where supported")
```
?

njhollinghurst: If we want the option to leave it unchanged, it can't just be a bool. Sadly, an optional bool would require extra processing.

PR #438, opened 2023-01-13
options: Automatically set H.264 level 4.2 if needed

naushir:
If the macroblocks/s exceed 245760, override the level option to "4.2".

### DIFF
--- a/core/video_options.hpp
+++ b/core/video_options.hpp
@@ -144,6 +144,14 @@ struct VideoOptions : public Options
 		if ((split || segment) && output.find('%') == std::string::npos)
 			LOG_ERROR("WARNING: expected % directive in output filename");
 
+		// From https://en.wikipedia.org/wiki/Advanced_Video_Coding#Levels
+		double mbps = ((width + 15) >> 4) * ((height + 15) >> 4) * framerate.value_or(DEFAULT_FRAMERATE);
+		if ((codec == "h264" || codec == "libav") && mbps > 245760.0)
+		{
+			LOG(1, "Overriding H.264 level 4.2");
+			level = "4.2";
+		}
+
 		return true;
 	}
 	virtual void Print() const override


core/video_options.hpp
```diff
@@ -144,6 +144,13 @@ struct VideoOptions : public Options
 		if ((split || segment) && output.find('%') == std::string::npos)
 			LOG_ERROR("WARNING: expected % directive in output filename");
 
+		// From https://en.wikipedia.org/wiki/Advanced_Video_Coding#Levels
+		if (width * height * framerate.value_or(DEFAULT_FRAMERATE) / (16 * 16) > 245760)
```
6by9, line 148: Rounding means that this is wrong.

1920x1080 is 120x68 = 8160 per frame (1080 not being a multiple of 16).
Using a bare 1920x1080 gives me a max frame rate of 30.34fps, when the actual limit is 30.11fps.

njhollinghurst: And only if level is not explicitly set by the user?

naushir: I made this explicit as if the user sets 4.1 with a rate that overflows the limit, it won't work. That brings us back to square one.

6by9: Seeing as level 4.2 is the max that we can realistically expect to achieve in real time, this seems reasonable.

It does mean the app will be a little noisy if --level 4.2 has been explicitly specified on the command line though.

naushir: > Rounding means that this is wrong.
> 
> 1920x1080 is 120x68 = 8160 per frame (1080 not being a multiple of 16). Using a bare 1920x1080 gives me a max frame rate of 30.34fps, when the actual limit is 30.11fps.

Given that framerate is a flot, rounding should be handled correctly right? Unless I'm missing something obvious...

naushir: > Seeing as level 4.2 is the max that we can realistically expect to achieve in real time, this seems reasonable.
> 
> It does mean the app will be a little noisy if --level 4.2 has been explicitly specified on the command line though.

I wonder if we should test the upper limit and bomb out as well?

6by9: > Given that framerate is a float, rounding should be handled correctly right? Unless I'm missing something obvious..

It's a macroblock limit, not pixel limit. A macroblock is 16x16 pixels and you can't encode fractional MBs.

Both width and height need to be rounded up to a multiple of 16 before multiplication.

naushir:
Fixed up the rounding, and added a conditional check on the codec type.

PR #441, opened 2023-01-18
build: Set a SONAME version in the libraries

martinezjavier:
The libraries built are currently unversioned, but donwstream users expect these to have a proper SONAME version in order to package them. Most Linux distributions packaging guidelines for example have this as a requirement.

There are alredy released versions and git tags, so let's use those as the libraries version, and the major number as the SONAME version. This is the convention used by most projects.

Signed-off-by: Javier Martinez Canillas <javierm@redhat.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.6)
 
-project(libcamera-apps)
+project(libcamera-apps VERSION 1.1.0)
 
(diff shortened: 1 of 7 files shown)


naushir:
@XECDesign, do you see any implications on our packaging flow with this change?

naushir:
@martinezjavier thanks for the patch.  I'll merge this once the tests have finished running!

PR #443, opened 2023-01-23
image: Add Subject Distance to EXIF data, to encode Lens Position

njhollinghurst:
This is the only bit of AF metadata that seems to match a standard Exif tag. [XXX Because of numerical/API differences, the JPEG and DNG versions can encode slightly different values.]

### DIFF
--- a/image/dng.cpp
+++ b/image/dng.cpp
@@ -5,6 +5,7 @@
  * dng.cpp - Save raw image as DNG file.
  */
 
+#include <limits>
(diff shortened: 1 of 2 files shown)


njhollinghurst:
Re-pushed with style fix.

Infinity seems to work. `exiftool` displays distances correctly in both JPG and DNG (though they don't always match beyond 4 significant figures). Raw-processing tools are not fazed by the new tag.

PR #453, opened 2023-02-02
core: minor code refactoring

chobitsfan:
We already have a refence to `stream->configuration()`. I think we can use it
https://github.com/raspberrypi/libcamera-apps/blob/40e001251d04a7e416ecfcfd883e48016c39e912/core/libcamera_app.cpp#L778


naushir:
Hi, thank you for this tidy-up.  Happy to merge this, but could you possibly update the commit message to something like:

```
core: Minor code refactoring

Reuse the existing StreamConfiguration reference in GetStreamInfo()
```

Also, please add a Signed-off-by: tag to the commit message (you can do this by ``git commit -s ...``).

chobitsfan:
Hi @naushir Thank you for reviewing and comments. It is done.

PR #458, opened 2023-02-07
encoder: H264: Trap framerate=0 to avoid division by zero error

njhollinghurst:
Without this change, `libcamera-vid -t 0 --framerate 0` would typically lock up after a few hundred frames.

Also, I've increased the denominator to 90kHz so that rates such as 30fps, 29.97fps can be represented accurately.

### DIFF
--- a/encoder/h264_encoder.cpp
+++ b/encoder/h264_encoder.cpp
@@ -131,12 +131,16 @@ H264Encoder::H264Encoder(VideoOptions const *options, StreamInfo const &info)
 	if (xioctl(fd_, VIDIOC_S_FMT, &fmt) < 0)
 		throw std::runtime_error("failed to set capture format");
 
-	struct v4l2_streamparm parm = {};
-	parm.type = V4L2_BUF_TYPE_VIDEO_OUTPUT_MPLANE;
-	parm.parm.output.timeperframe.numerator = 1000 / options->framerate.value_or(DEFAULT_FRAMERATE);
-	parm.parm.output.timeperframe.denominator = 1000;
-	if (xioctl(fd_, VIDIOC_S_PARM, &parm) < 0)
-		throw std::runtime_error("failed to set streamparm");
+	double frate = options->framerate.value_or(DEFAULT_FRAMERATE);
+	if (frate > 0.0)
+	{
+		struct v4l2_streamparm parm = {};
+		parm.type = V4L2_BUF_TYPE_VIDEO_OUTPUT_MPLANE;
+		parm.parm.output.timeperframe.numerator = 90000.0 / frate;
+		parm.parm.output.timeperframe.denominator = 90000;
+		if (xioctl(fd_, VIDIOC_S_PARM, &parm) < 0)
+			throw std::runtime_error("failed to set streamparm");
+	}
 
 	// Request that the necessary buffers are allocated. The output queue
 	// (input to the encoder) shares buffers from our caller, these must be


6by9:
> Without this change, libcamera-vid -t 0 --framerate 0 would typically lock up after a few hundred frames.

Just omitting the framerate means that it'll be using the default of 30, and rate control will be based on that.

Framerate 0 should use the timestamps for rate control, but that does mean that providing them (and in the correct units) becomes mandatory.

> Also, I've increased the denominator to 90kHz so that rates such as 30fps, 29.97fps can be represented accurately.

The H264 codec itself only handles integers for passing into the rate control algorithm, and also in the SEI timing header.

njhollinghurst:
I can set a zero if that is preferable.  Which flavour of zero is recommended?
- timeperframe.numerator = 0, timeperframe.denominator = 0
- timeperframe.numerator = 0, timeperframe.denominator = 1
- timeperframe.numerator = 1, timeperframe.denominator = 0

6by9:
Oh V4L2's stupid use of frame interval!

This is unlikely to be portable, but such is life.

bcm2835-v4l2-codec.c vidioc_s_parm takes
```
	ctx->framerate_num =
			parm->parm.output.timeperframe.denominator;
	ctx->framerate_denom =
			parm->parm.output.timeperframe.numerator;
```
and is passed into the MMAL port format as
```
		port->es.video.frame_rate.num = ctx->framerate_num;
		port->es.video.frame_rate.den = ctx->framerate_denom;
```
MMAL will be looking for video.frame_rate.num = 0 if we're wanting the codec to see a frame rate of 0.

```timeperframe.numerator = 1, timeperframe.denominator = 0``` should therefore end up with an appropriate configuration.

njhollinghurst:
Tangentially, the crashing condition was:
- timeperframe.numerator = UINT_MAX, timeperframe.denominator = 1000
where UINT_MAX was the result of casting NaN. I hope none of those other combinations cause a crash!

njhollinghurst:
> bcm2835-v4l2-codec.c vidioc_s_parm takes
> 
> ```
> 	ctx->framerate_num =
> 			parm->parm.output.timeperframe.denominator;
> 	ctx->framerate_denom =
> 			parm->parm.output.timeperframe.numerator;
> ```
> 
> and is passed into the MMAL port format as
> 
> ```
> 		port->es.video.frame_rate.num = ctx->framerate_num;
> 		port->es.video.frame_rate.den = ctx->framerate_denom;
> ```
> 
> MMAL will be looking for video.frame_rate.num = 0 if we're wanting the codec to see a frame rate of 0.
> 
> `timeperframe.numerator = 1, timeperframe.denominator = 0` should therefore end up with an appropriate configuration.

Just above the lines mentioned, it rejects cases where either the numerator or denominator is zero (returning `-EINVAL`). So it looks like a frame rate of 0 cannot be signalled.

In practice, when libcamera-vid has `framerate=0` it seems to run at 30fps most of the time. We could either
- quietly assume that to be the default (as committed)
- or signal libcamera-apps' `DEFAULT_FRAMERATE` explicitly?

njhollinghurst:
The ioctl is sent from the Encoder constructor, before libcamera is initialized, so without rearrangement there is no access to frame duration limits.

Agree that the proper fix is to make the codec tolerate zero or infinite durations (we should support whichever is correct from a V4L2 perspective, and maybe the other zero combinations as well).

libcamera-apps PR withdrawn for now.

njhollinghurst:
@6by9 and I had a look at V4L2 drivers and v4l2-compliance: the whole thing is clear as mud, but it _appears_ that zeros anywhere in timeperframe should get sanitized to nonzeros, and zero can't be used to signal a variable frame rate.

I'm going to reopen this PR as is: it avoids the lockup, and (by luck) it defaults to a reasonable frame rate (30fps).

PR #465, opened 2023-02-11
Install library headers

knro:
Install library headers to allow 3rd party code to link and build against the libraries provided by libcamera-apps.

This is used for [INDI Libcamera](https://github.com/indilib/indi-3rdparty/tree/master/indi-libcamera) driver but would be useful for any 3rd party project that wants to build on libcamera-apps.


naushir:
@krno thank you for this submission.  I think the change is useful, and we should include it.  Would you be able to re-submit this with a commit message that better follows our convention?  Something like:

```
build: Install library headers

Install the library headers into the system install location...
```

Please also do add a ``Signed-off-by:`` tag to the end of the commit message.

marcusfolkesson:
@knro do you intend to update the submission? Otherwise, I can take over.

knro:
Please go ahead I got so busy and didn't have time to update it.

PR #476, opened 2023-02-22
options: Set the custom tuning file as early as possible

naushir:
Currently the LIBCAMERA_RPI_TUNING_FILE is set after the --list-cameras
option handling.  This causes the camera to not enumerate if a standard
tuning file is unavailable, and --list-cameras does not list it.

Move setting LIBCAMERA_RPI_TUNING_FILE to earlier in the options handler
function to fix this.


naushir:
@davidplowman @njhollinghurst

davidplowman:
LGTM!

PR #482, opened 2023-02-28
image: dng: Add support for 16-bit Bayer formats

njhollinghurst:
This code assumes that SRGGB16 etc are stored in memory in _native_ byte order (which is what TIFFWriteScanLine expects), therefore the "conversion" from 16 to 16 bits can just be a memcpy.


njhollinghurst:
The second commit fixes unnecessarily dark thumbnails. The previous 8-bit normalization didn't account for the square root, and ended up with only 7 significant bits.

PR #485, opened 2023-03-09
build: install library headers

marcusfolkesson:
To let 3rd party code to link and and build against libcamera-apps we also have to provide the header files.

Install the library headers into the system as well.

Credit to Jasem Mutlaq who is the original author of this patch.


naushir:
Thanks, this is ready to merge now!

PR #500, opened 2023-04-06
Codec changes

naushir:


encoder/encoder.cpp
```diff
@@ -16,9 +17,28 @@
...
+		return new H264Encoder(options, info);
+
+#if LIBAV_PRESENT
+	// No hardware codec available, use x264 through libav.
+	options->libav_video_codec = "libx264";
+	return new LibAvEncoder(options, info);
+#endif
+
+	return new MjpegEncoder(options);
+}
```
davidplowman, line 35: Is the libav codec available by default these days? I'm not sure about falling back to MJPEG just because you don't have a /dev/video11?

encoder/libav_encoder.cpp
```diff
@@ -26,14 +28,41 @@ void encoderOptionsH264M2M(VideoOptions const *options, AVCodecContext *codec)
...
 	{ "h264_v4l2m2m", encoderOptionsH264M2M },
+	{ "libx264", encoderOptionsLibx264 },
 };
 
+} // namespace
+
 void LibAvEncoder::initVideoCodec(VideoOptions const *options, StreamInfo const &info)
 {
-	const std::string codec_name("h264_v4l2m2m");
+	const std::string codec_name(options->libav_video_codec);
```
davidplowman, line 65: Just wondering very slightly whether the `codec_name` variable is even necessary? Not that it matters...

encoder/libav_encoder.cpp
```diff
@@ -26,14 +28,41 @@ void encoderOptionsH264M2M(VideoOptions const *options, AVCodecContext *codec)
 		codec->bit_rate = options->bitrate;
 }
 
-static const std::map<std::string, std::function<void(VideoOptions const *, AVCodecContext *)>> optionsMap =
+void encoderOptionsLibx264(VideoOptions const *options, AVCodecContext *codec)
+{
+	codec->pix_fmt = AV_PIX_FMT_YUV420P;
+
+	if (options->bitrate)
+		codec->bit_rate = options->bitrate * 1000;
```
davidplowman, line 36: I thought our `options->bitrate` is bits per second (to match raspivid). Does that make `codec->bit_rate` millibits per second? Or am I hallucinating??

encoder/encoder.cpp
```diff
@@ -16,9 +17,28 @@
...
 
-Encoder *Encoder::Create(VideoOptions const *options, const StreamInfo &info)
+Encoder *auto_codec_select(VideoOptions *options, const StreamInfo &info)
 {
-	if (strcasecmp(options->codec.c_str(), "yuv420") == 0)
+	const char hw_codec[] = "/dev/video11";
+	struct stat buffer;
+
+	if (!stat(hw_codec, &buffer))
+		return new H264Encoder(options, info);
```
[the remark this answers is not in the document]
naushir: Yes, perhaps i should open the file and query the device properties?

encoder/encoder.cpp
naushir: On the full fat builds it is.  Not sure what the best fallback without HW codec or libav would be?  Perhaps plain YUV?

encoder/libav_encoder.cpp
naushir: I can remove that.

encoder/libav_encoder.cpp
naushir: > Does that make `codec->bit_rate` millibits per second? Or am I hallucinating??

Yes it does, but only for the x264 case.  This is the units it requires.

naushir:
@davidplowman and @njhollinghurst if there are no further comments, I'll merge this in.

davidplowman:
Go for it! I assume the style complaints are ones we want to ignore...

naushir:
Yup, we can ignore the style warnings for this one.  Merging!

PR #504, opened 2023-05-04
libcamera_app: Fix cmake RPATH stripping when installing libcamera_ap…

naushir:
…p.so

With recent versions of libcamera, it seems that the default behaviour of cmake stripping the runtime path from the .so file causes a failure.

Fix this by ensuring the RPATH does not get stripped.

For further details, see:
https://stackoverflow.com/questions/32469953/why-is-cmake-designed-so-that-it-removes-runtime-path-when-installing


naushir:
@davidplowman could you give this a try and see if this fix works for you?

PR #506, opened 2023-05-10
New meson build system

naushir:


post_processing_stages/meson.build
```diff
@@ -0,0 +1,49 @@
...
+        'object_detect_draw_cv_stage.cpp',
+    ])
+    libcamera_app_dep += opencv_dep
+else
+    enable_opencv = false
+endif
+
+enable_tflite = get_option('enable_tflite')
+tflite_dep = dependency('tensorflow-lite', required : false)
+if enable_tflite and tflite_dep.found()
```
eli-schwartz, line 36: Maybe you want to use a feature option here?

```meson
tflite_dep = dependency('tensorflow-lite', required : get_option('enable_tflite'))
if tflite_dep.found()
    ...
```

You can then pass `-Denable_tflite=disabled` and meson will log that tensorflow-lite was not searched for because the feature "enable_tflite" was disabled. It will simply return not-found without running pkg-config.

(You might want to change the name from enable_tflite to just "tflite" in that case.)

PR #507, opened 2023-05-25
Fix CI

wh201906:
In the commit 726e9274ea95fa46352556d340c5793a8da51fcd of libcamera, the -Dpipelines option has changed from "raspberrypi" to "rpi/vc4". You can check this change in the commit message.

https://git.libcamera.org/libcamera/libcamera.git/commit/?id=726e9274ea95fa46352556d340c5793a8da51fcd


naushir:
Thank you for this fix.  The change itself is good, but would you be able to re-submit with a modified commit message that follows our standard format.  you can see examples in the various commits in the git log.  Please also add a Singed-off-by: tag.

PR #518, opened 2023-06-13
Set the pipeline configuration file

davidplowman:
If the LIBCAMERA_RPI_CONFIG_FILE environment variable is set then we respect it, otherwise we point it at the /usr/local/share or /usr/share version of rpi_apps.yaml, whichever we find first.

### DIFF
--- a/core/libcamera_app.cpp
+++ b/core/libcamera_app.cpp
@@ -13,8 +13,10 @@
 
 #include <cmath>
 #include <fcntl.h>
+#include <stdlib.h>
 
 #include <sys/ioctl.h>
+#include <sys/stat.h>
 
 #include <linux/videodev2.h>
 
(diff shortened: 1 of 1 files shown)


core/libcamera_app.cpp
```diff
@@ -64,13 +66,40 @@ static libcamera::PixelFormat mode_to_pixel_format(Mode const &mode)
...
+		"/usr/local/share/libcamera/pipeline/rpi/vc4/rpi_apps.yaml",
+		"/usr/share/libcamera/pipeline/rpi/vc4/rpi_apps.yaml",
+	};
+
+	for (auto &config_file : config_files)
+	{
+		struct stat info;
+		if (stat(config_file.c_str(), &info) == 0)
+		{
+			printf("Setting %s\n", config_file.c_str());
```
naushir, line 87: LOG() instead of printf?

PR #526, opened 2023-06-28
tests: Update shutter test to use 2x gain

naushir:
The check_jpeg_shutter test would fail on the imx708 as the minimum gain is 1.12.  Update the test to use a gain of 2x to get around this.


naushir:
@davidplowman this wants to go in before the next release.

Issue #527, opened 2023-07-01
API to get at CameraManager or Camera list


anjok:
For the INDI libcamera driver, we need to get at the camera list before actually opening a camera. Also, we need to be able to Open and Close the camera on a regular basis. Unfortunately, the API that is present needs the camera manager as input.

Can you add a means for that to get the list? Currently, I cloned the code into the driver and  added a method to get a pointer to CameraManager and run the same code that you run for discovering cameras:

https://github.com/indilib/indi-3rdparty/blob/master/indi-libcamera/libcamera-apps/core/libcamera_app.hpp#L118
https://github.com/indilib/indi-3rdparty/blob/master/indi-libcamera/indi_libcamera.cpp#L595C22-L595C22

I can't reuse the one you added to LibCameraApp::GetCameraList() as it takes a std::unique_pointer argument and I can't return the one you created for the instance.

This is of course not very usable, as you'll need to recompile this each time something changes.

naushir:
Hi, unfortunately it's a bit hard for me to guess exactly what API changes you need here without more context.   If you are able to create a basic PR with the changes, we would be happy to discuss this in more detail and merge the appropriate changes.

Vulisha:
Hi, let me jump in with probable guess, anjok may confirm if my suspicions are true. 

In INDI we do not know what is the camera is used as many may be used, and INDI is. made as universal driver so we need to get info what model of camera is connected and its properties before we Open the camera, and in current api we need to know Camera before we can start it and use GetCameraList.

naushir:
The only way to enumerate cameras in libcamera is through the camera manager.

Again, I am a bit unclear on how you want to do things in your INDI framework - I've never used it before. If you could create a PR with some proposed API changes that would work for you we can get something suitable merged for your. The PR does not have to be functionally complete to start with.

anjok:
Here you go:

https://github.com/raspberrypi/libcamera-apps/pull/528/files

I  have no idea if this is how std:unique_ptr is supposed to get used, though. Since we have an instance of LibcameraApp in the driver, we might as well add an instance method 

	std::vector<std::shared_ptr<libcamera::Camera>> GetCameras();

which calls the static method with it's CameraManager, but my std:: C++ skills aren't that up to speed.

anjok:
The PR above is what I added to out local copy of the source files. If you'd add something like this, we don'T have to maintain that copy anymore (provided the user has whatever version of the libcamera-apps lib then contains it)

[...]

PR #528, opened 2023-07-02
core: adding method to access CameraManager

anjok:
one way to fix issue https://github.com/raspberrypi/libcamera-apps/issues/527

### DIFF
--- a/core/libcamera_app.cpp
+++ b/core/libcamera_app.cpp
@@ -136,7 +136,7 @@ void LibcameraApp::OpenCamera()
 	if (ret)
 		throw std::runtime_error("camera manager failed to start, code " + std::to_string(-ret));
 
-	std::vector<std::shared_ptr<libcamera::Camera>> cameras = LibcameraApp::GetCameras(camera_manager_);
+	std::vector<std::shared_ptr<libcamera::Camera>> cameras = GetCameras();
 	if (cameras.size() == 0)
 		throw std::runtime_error("no cameras available");
 	if (options_->camera >= cameras.size())
@@ -771,6 +771,11 @@ libcamera::Stream *LibcameraApp::GetMainStream() const
 	}
 
 	return nullptr;
+}
+
+const libcamera::CameraManager *LibcameraApp::GetCameraManager() const
+{
+	return camera_manager_.get();
 }
 
 std::vector<libcamera::Span<uint8_t>> LibcameraApp::Mmap(FrameBuffer *buffer) const
--- a/core/libcamera_app.hpp
+++ b/core/libcamera_app.hpp
@@ -119,6 +119,12 @@ public:
 	Stream *LoresStream(StreamInfo *info = nullptr) const;
 	Stream *GetMainStream() const;
 
+	const CameraManager *GetCameraManager() const;
+	std::vector<std::shared_ptr<libcamera::Camera>> GetCameras()
+	{
+		return GetCameras(camera_manager_.get());
+	}
+
 	std::vector<libcamera::Span<uint8_t>> Mmap(FrameBuffer *buffer) const;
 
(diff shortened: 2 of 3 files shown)


core/libcamera_app.cpp
```diff
@@ -773,6 +773,11 @@ libcamera::Stream *LibcameraApp::GetMainStream() const
 	return nullptr;
 }
 
+libcamera::CameraManager *LibcameraApp::GetCameraManager() const
```
naushir, line 776: Perhaps this function should return a ``const libcamera::CameraManager *`` value?

naushir (commented):
This changes seems reasonable, so I am happy to merge something along these lines.  However, from our discussion on https://github.com/raspberrypi/libcamera-apps/issues/527, you will want to pass this pointer into ``LibcameraApp::GetCameras()`` which takes in a ``const std::unique_ptr<CameraManager> &`` as a parameter.  Perhaps this function also needs to change to take in a ``const libcamera::CameraManager *`` instead?

naushir:
One more thing, could you add a small description to the commit message, and end it with a ``Signed-off-by:`` tag?

core/libcamera_app.cpp
anjok: sure. as I said, my c++ is stuck in the late 90s...

naushir: This ought to change to return ``const libcamera::CameraManager *``

core/libcamera_app.hpp
```diff
@@ -119,6 +119,12 @@ class LibcameraApp
 	Stream *LoresStream(StreamInfo *info = nullptr) const;
 	Stream *GetMainStream() const;
 
+	CameraManager *GetCameraManager() const;
+	std::vector<std::shared_ptr<libcamera::Camera>> GetCameras()
+	{
+		return GetCameras(camera_manager_);
+	}
+
```
naushir, line 127: I don't think the ``GetCameras()`` override is needed right?  Instead you need to change (below):
```
static std::vector<std::shared_ptr<libcamera::Camera>> GetCameras(const std::unique_ptr<CameraManager> &cm)
```
to
```
static std::vector<std::shared_ptr<libcamera::Camera>> GetCameras(const CameraManager> *cm)
```

naushir (commented):
Additionally, in the commit message, could you please use your name and email on the Signed-off-by line?

```
Signed-off-by: Your Name <your@email.address>
```

core/libcamera_app.hpp
anjok: right. 

But I already have a LibCameraApp, so I might as well just use a new app->GetCameras() off of it and not deal with the camera manager at all? After all, I was only doing that because that's what the multiple other code pieces in LibCameraApp did and you didn't have a singular mens to get at the list.

Also, you'd need to change "public", if interim, API.. I don't care one way or another, but given there's no way to get at the manager without an app in the first place, and there's not way to get a unique_ptr without having one first, I would recommend (again without knowing a lot about what all of this does or should do) to remove the static method and just have an instance method....

anjok: OK, so the options use that method... why? I guess this is where our problems came from, as we're using the options when the app has already been created. So now you have two cam managers... at least I think so.

anjok:
> Additionally, in the commit message, could you please use your name and email on the Signed-off-by line?
> 
> ```
> Signed-off-by: Your Name <your@email.address>
> ```

Fixed... but I was under the impression that this would be the committer who said "this is ok to merge"? Why can I sign off my own stuff as an external?

core/libcamera_app.hpp
naushir: There is another place that uses GetCameras() with the camera manager already opened.  This is why passing in an external camera manager must still be supported.

PR #535, opened 2023-07-14
Few changes to the time based option handling.

naushir:


naushir:
@davidplowman would you be able to eyeball these please?

PR #537, opened 2023-07-17
encoder: New LibAV segment, split and pause implementation

bebon901:
Reprogrammed the implementation of segment, split and pause for when encoding with LibAV. Moved functionality to the encode thread New version should make it easier to implement timestamping and metadata for files.

### DIFF
--- a/encoder/libav_encoder.cpp
+++ b/encoder/libav_encoder.cpp
@@ -65,7 +65,22 @@ const std::map<std::string, std::function<void(VideoOptions const *, AVCodecCont
 
 void LibAvEncoder::Signal()
 {
-	feed_encoder_frames_ = !feed_encoder_frames_;
+	// We now need to tell the rest of the program what is happening
+	pause_state_.paused = !pause_state_.paused;
+	if (pause_state_.first_pause)
+		pause_state_.first_pause = false;
+	// When we switch from paused to not paused, update the timestamp at which that happened
+	if (!pause_state_.paused)
+	{
+		segment_start_ts_ = current_timestamp_;
+		pause_state_.unpause_timestamp = current_timestamp_ - pause_state_.pause_duration - video_start_ts_;
+	}
+	// Unpaused to Pause means that we have paused. Similarly update timestamps following
+	if (pause_state_.paused)
+	{
+		pause_state_.pause_timestamp = current_timestamp_ - pause_state_.pause_duration - video_start_ts_;
+		pause_state_.unpause_timestamp = 0;
+	}
 }
 
 void LibAvEncoder::initVideoCodec(VideoOptions const *options, StreamInfo const &info)
@@ -279,8 +294,7 @@ void LibAvEncoder::initAudioOutCodec(VideoOptions const *options, StreamInfo con
 LibAvEncoder::LibAvEncoder(VideoOptions const *options, StreamInfo const &info)
 	: Encoder(options), output_ready_(false), abort_video_(false), abort_audio_(false), video_start_ts_(0),
 	  audio_samples_(0), in_fmt_ctx_(nullptr), out_fmt_ctx_(nullptr), segment_start_ts_(0), segment_num_(0),
-	  virtual_video_ts_(0), virtual_audio_ts_(0), previous_video_timestamp_(0), previous_audio_timestamp_(0),
-	  feed_encoder_frames_(true), previous_feed_value_(true)
+	  previous_video_timestamp_(0), current_timestamp_(0)
 {
 	avdevice_register_all();
 
@@ -339,14 +353,12 @@ void LibAvEncoder::EncodeBuffer(int fd, size_t size, void *mem, StreamInfo const
 	{
 		video_start_ts_ = timestamp_us;
 		segment_start_ts_ = video_start_ts_;
+		pause_state_.pause_timestamp = 0;
+		pause_state_.unpause_timestamp = 0;
 	}
-	// Segmentation Function
-	if ((options_->segment && (timestamp_us - segment_start_ts_) / 1000 > options_->segment) && output_ready_)
-		nextSegment(timestamp_us, segment_start_ts_);
-	// Split Function
-	//want to catch the 'rising edge' of the feed_frames variable. when it rises, we want to save the file and start with a new one
-	if (options_->split && previous_feed_value_ != feed_encoder_frames_ && feed_encoder_frames_)
-		nextSegment(timestamp_us, virtual_video_ts_);
+	current_timestamp_ = timestamp_us + (options_->av_sync.value < 0us
+											 ? -options_->av_sync.get<std::chrono::microseconds>()
+											 : 0); // Standard Recording modes
 
 	frame->format = codec_ctx_[Video]->pix_fmt;
 	frame->width = info.width;
@@ -356,21 +368,7 @@ void LibAvEncoder::EncodeBuffer(int fd, size_t size, void *mem, StreamInfo const
 	if (previous_video_timestamp_ == 0)
 		previous_video_timestamp_ = timestamp_us; // Only useful on startup
 
-	int64_t elapsed_time = timestamp_us - previous_video_timestamp_; // Time since last loop. Needed for pause function
-
-	if (feed_encoder_frames_ && options_->keypress)
-	{
-		// If feeding video and keypress option is on, then count the frame timer for the frames
-		virtual_video_ts_ += elapsed_time;
-		frame->pts = virtual_video_ts_; // Gives the frames a relative time since recording started
-	}
-	else
-	{
-		frame->pts = timestamp_us - video_start_ts_ +
-				(options_->av_sync.value < 0us ? -options_->av_sync.get<std::chrono::microseconds>() : 0); // Standard Recording modes
-	}
-	previous_video_timestamp_ = timestamp_us;
-	previous_feed_value_ = feed_encoder_frames_;
+	frame->pts = current_timestamp_ - video_start_ts_;
 
 	if (codec_ctx_[Video]->pix_fmt == AV_PIX_FMT_DRM_PRIME)
 	{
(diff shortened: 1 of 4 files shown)


encoder/libav_encoder.cpp
```diff
@@ -723,30 +799,13 @@ void LibAvEncoder::audioThread()
 			AVRational num = { 1, out_frame->sample_rate };
 			int64_t ts = av_rescale_q(audio_samples_, num, codec_ctx_[AudioOut]->time_base);
 
-			// Need to set up timestamps based on the video recording settings:
-			if (previous_audio_timestamp_ == 0)
-				previous_audio_timestamp_ = ts; // Only useful on startup
-
-			int64_t et = ts - previous_audio_timestamp_; // Elapsed Time
+			out_frame->pts = ts + (options_->av_sync < 0 ? -options_->av_sync : 0);
```
naushir, line 802: Have a look at https://github.com/raspberrypi/libcamera-apps/commit/f5fe27124531af19f68dd8ead18a05773c0f1db3.

I think this wants to be
```
out_frame->pts = ts + (options_->av_sync > 0 ? options_->av_sync : 0);
```

encoder/libav_encoder.cpp
```diff
@@ -279,8 +292,8 @@ void LibAvEncoder::initAudioOutCodec(VideoOptions const *options, StreamInfo con
 LibAvEncoder::LibAvEncoder(VideoOptions const *options, StreamInfo const &info)
 	: Encoder(options), output_ready_(false), abort_video_(false), abort_audio_(false), video_start_ts_(0),
 	  audio_samples_(0), in_fmt_ctx_(nullptr), out_fmt_ctx_(nullptr), segment_start_ts_(0), segment_num_(0),
-	  virtual_video_ts_(0), virtual_audio_ts_(0), previous_video_timestamp_(0), previous_audio_timestamp_(0),
-	  feed_encoder_frames_(true), previous_feed_value_(true)
+	  previous_video_timestamp_(0), pause_timestamp_(-100000), unpause_timestamp_(-1000000), pausestate_(false),
```
naushir, line 295: Why -100000?  I think 0 would be preferred to be consistent with the rest of the init values.

encoder/libav_encoder.cpp
```diff
@@ -356,21 +363,7 @@ void LibAvEncoder::EncodeBuffer(int fd, size_t size, void *mem, StreamInfo const
...
-		frame->pts = virtual_video_ts_; // Gives the frames a relative time since recording started
-	}
-	else
-	{
-		frame->pts = timestamp_us - video_start_ts_ +
-					 (options_->av_sync < 0 ? -options_->av_sync : 0); // Standard Recording modes
-	}
-	previous_video_timestamp_ = timestamp_us;
-	previous_feed_value_ = feed_encoder_frames_;
+	frame->pts = timestamp_us - video_start_ts_ + (options_->av_sync < 0 ? -options_->av_sync : 0);
```
naushir, line 366: Can this be simplified to
```c++
frame->pts = current_timestamp_ - video_start_ts_;
```
?

encoder/libav_encoder.cpp
```diff
@@ -499,27 +492,101 @@ void LibAvEncoder::encode(AVPacket *pkt, unsigned int stream_id)
...
+			pause_duration_ += current_timestamp_ - previous_video_timestamp_;
+		previous_video_timestamp_ = current_timestamp_;
 		std::scoped_lock<std::mutex> lock(output_mutex_);
-		// pkt is now blank (av_interleaved_write_frame() takes ownership of
-		// its contents and resets pkt), so that no unreferencing is necessary.
-		// This would be different if one used av_write_frame().
-		ret = av_interleaved_write_frame(out_fmt_ctx_, pkt);
+
+		// We need a copy of what the packet timestamp was
+		int64_t packet_ts_ = pkt->pts;
```
naushir, line 502: ``packet_ts_`` should probably be renamed to ``packet_ts``.

encoder/libav_encoder.cpp
```diff
@@ -499,27 +492,101 @@ void LibAvEncoder::encode(AVPacket *pkt, unsigned int stream_id)
...
+
+		/*
+		Need to detect what state the packet came from
+		If the timestamp is after a pause timestamp and before an unpause ts then it must be paused, and hence thrown away
+		Additionally, the timestamp can be after a pause event but the current unpause timestamp is equal to zero, and hence is also paused
+		if the timestamp is after an unpause timestamp and after the latest pause timestamp then it must be written (unpaused state)
+
+		In the following, the logic all happens when we are passed a video frame. i.e. starting a new segment, initating split.
+		Audio just follows a through afterwards. This is needed to make sure that the first thing in the video file is a keyframe.
+		*/
```
naushir, line 512: Better to use C++ style comments (``\\``) for this block.

encoder/libav_encoder.cpp
```diff
@@ -499,27 +492,101 @@ void LibAvEncoder::encode(AVPacket *pkt, unsigned int stream_id)
...
+					// This is the first frame that is a keyframe. This will be the start of a new segment/split.
+					ret = av_interleaved_write_frame(out_fmt_ctx_, pkt);
+					first_frame_ = false;
+					write_frames_ = true;
+				}
+				// If there's nothing special going on, just write the frame
+				else if (!first_frame_ && write_frames_)
+					ret = av_interleaved_write_frame(out_fmt_ctx_, pkt);
+			}
+		}
```
naushir, line 546: I think this probably wants to go into a separate function handler.

encoder/libav_encoder.cpp
```diff
@@ -499,27 +492,101 @@ void LibAvEncoder::encode(AVPacket *pkt, unsigned int stream_id)
...
+				// This is the first keyframe that has arrived since the start of a segment
+				first_frame_ = false;
+				segment_start_ts_ = current_timestamp_;
+				ret = av_interleaved_write_frame(out_fmt_ctx_, pkt);
+				write_frames_ = true;
+			}
+			else // Again, not much going on here, just write the frame
+				ret = av_interleaved_write_frame(out_fmt_ctx_, pkt);
+		}
+
```
naushir, line 569: As above, this probably wants to go into a separate function handler.

encoder/libav_encoder.cpp
```diff
@@ -499,27 +492,101 @@ void LibAvEncoder::encode(AVPacket *pkt, unsigned int stream_id)
...
+					first_frame_ = false;
+					write_frames_ = true;
+				}
+				// If there's nothing special going on, just write the frame
+				else if (!first_frame_ && write_frames_)
+					ret = av_interleaved_write_frame(out_fmt_ctx_, pkt);
+			}
+		}
+
+		if (options_->segment)
```
naushir, line 548: ``options_->segment`` probably wants to now became our new fancy ``TimeVal`` type so users can specify units on the command line.

davidplowman:
> Conflicts wise, I think we just want to overwrite the old file. But will check.

I wonder if one of Naush's recent changes has caused a conflict. Maybe rebase the latest main into your local tree, sort out any conflicts, and then re-push.

encoder/libav_encoder.cpp
```diff
@@ -65,7 +65,20 @@ const std::map<std::string, std::function<void(VideoOptions const *, AVCodecCont
 
 void LibAvEncoder::Signal()
 {
-	feed_encoder_frames_ = !feed_encoder_frames_;
+	// We now need to tell the rest of the program what is happening
+	pausestate_ = !pausestate_;
```
davidplowman, line 69: Our convention in libcamera_apps is normally to put an underscore in variable names to break words, so it would be `paused_state`. But see my other comment below!

encoder/libav_encoder.cpp
```diff
@@ -279,8 +292,8 @@ void LibAvEncoder::initAudioOutCodec(VideoOptions const *options, StreamInfo con
 LibAvEncoder::LibAvEncoder(VideoOptions const *options, StreamInfo const &info)
 	: Encoder(options), output_ready_(false), abort_video_(false), abort_audio_(false), video_start_ts_(0),
 	  audio_samples_(0), in_fmt_ctx_(nullptr), out_fmt_ctx_(nullptr), segment_start_ts_(0), segment_num_(0),
-	  virtual_video_ts_(0), virtual_audio_ts_(0), previous_video_timestamp_(0), previous_audio_timestamp_(0),
-	  feed_encoder_frames_(true), previous_feed_value_(true)
+	  previous_video_timestamp_(0), pause_timestamp_(-100000), unpause_timestamp_(-1000000), pausestate_(false),
+	  previous_pausestate_(false), current_timestamp_(0), first_frame_(false), write_frames_(true), pause_duration_(0)
```
davidplowman, line 296: There are quite a few variables relating to pausing/segmenting. I wonder a bit about gathering them into structures so that it's clear which values relate to what. For example, we'd have a `PauseState` structure containing the variables relating to being paused (or not). The variable previously know as `pausestate_` might then be a field within this, such as `pause_state_.paused`.

I'm only asking the question here... what do folks think?

encoder/libav_encoder.cpp
```diff
@@ -499,27 +492,101 @@ void LibAvEncoder::encode(AVPacket *pkt, unsigned int stream_id)
...
+		Need to detect what state the packet came from
+		If the timestamp is after a pause timestamp and before an unpause ts then it must be paused, and hence thrown away
+		Additionally, the timestamp can be after a pause event but the current unpause timestamp is equal to zero, and hence is also paused
+		if the timestamp is after an unpause timestamp and after the latest pause timestamp then it must be written (unpaused state)
+
+		In the following, the logic all happens when we are passed a video frame. i.e. starting a new segment, initating split.
+		Audio just follows a through afterwards. This is needed to make sure that the first thing in the video file is a keyframe.
+		*/
+
+		if (options_->keypress)
```
davidplowman, line 514: You can do pausing with Linux signals as well as key presses. Do we need  to check for `options_->signal` too?

encoder/libav_encoder.cpp
naushir: Sounds like a good idea.  It will make the code tidier.

encoder/libav_encoder.cpp
```diff
@@ -499,27 +492,101 @@ void LibAvEncoder::encode(AVPacket *pkt, unsigned int stream_id)
...
+				else if (!first_frame_ && write_frames_)
+					ret = av_interleaved_write_frame(out_fmt_ctx_, pkt);
+			}
+		}
+
+		if (options_->segment)
+		{
+			if ((current_timestamp_ - segment_start_ts_) / 1000 > options_->segment && !first_frame_ &&
+				stream_id == Video)
+			{
```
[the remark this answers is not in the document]
bebon901: I can do this if wanted, I think it'll involve quite a few variable changes though? Might get annoying switching between the current ` 'int64_t'` and `'std::chrono::time_point`' though. Let me know if you want this added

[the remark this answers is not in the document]
naushir: No, let's not bother doing this right now and save it for another day.  But I do think switching ``options_->segment`` to a ``TimeVal`` would be nice...

davidplowman:
I think I'm OK to merge at this point, do you want one more look @naushir ?

naushir:
Looks good, let's merge it.

PR #538, opened 2023-07-18
Changes for bitrate units.

naushir:
Also a small commit to fix x64 builds with OpenCV.

### DIFF
--- a/core/video_options.hpp
+++ b/core/video_options.hpp
@@ -13,6 +13,70 @@
 
 #include "options.hpp"
 
+struct Bitrate
+{
+public:
+	Bitrate() : bps_(0) {}
+
+	void set(const std::string &s)
+	{
+		static const std::map<std::string, uint64_t> match
+		{
+			{ "bps", 1 },
+			{ "b", 1 },
+			{ "kbps", 1000 },
+			{ "k", 1000 },
+			{ "K", 1000 },
+			{ "mbps", 1000 * 1000 },
+			{ "m", 1000 * 1000 },
+			{ "M", 1000 },
+		};
+
+		try
+		{
+			std::size_t end_pos;
+			float f = std::stof(s, &end_pos);
+			bps_ = f;
+
+			for (const auto &m : match)
+			{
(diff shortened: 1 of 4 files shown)


naushir:
@davidplowman

core/video_options.hpp
```diff
@@ -13,6 +13,68 @@
...
+
+	void set(const std::string &s)
+	{
+		static const std::map<std::string, uint64_t> match
+		{
+			{ "bps", 1 },
+			{ "b", 1 },
+			{ "kbps", 1000 },
+			{ "k", 1000 },
+			{ "mbps", 1000 * 1000 },
```
6by9, line 29: Wait for the comment over supporting milli bits/sec.... ;-)

naushir: Haha, there was a whole conversation about it this morning :)

core/video_options.hpp
```diff
@@ -13,6 +13,68 @@
...
+			std::size_t end_pos;
+			float f = std::stof(s, &end_pos);
+			bps_ = f;
+
+			for (const auto &m : match)
+			{
+				auto found = s.find(m.first, end_pos);
+				if (found != end_pos || found + m.first.length() != s.length())
+					continue;
+				bps_ *= m.second;
```
davidplowman, line 44: What happens here if someone wants, for example, "0.5m". Are we going to round off the 0.5 and leave them with zero? Perhaps something like `bps_ = f  * m.second;` would be better?

core/video_options.hpp
```diff
@@ -13,6 +13,68 @@
...
+	void set(const std::string &s)
+	{
+		static const std::map<std::string, uint64_t> match
+		{
+			{ "bps", 1 },
+			{ "b", 1 },
+			{ "kbps", 1000 },
+			{ "k", 1000 },
+			{ "mbps", 1000 * 1000 },
+			{ "m", 1000 * 1000 },
```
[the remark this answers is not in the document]
naushir: Will do.

PR #540, opened 2023-07-19
Metadata and Timestamp

bebon901:
Have now added the ability to:
- Use metadata and timestamping when encoding with LibAV
- Save metadata and timestamp files to separate files when using split or segment options

Ready for review @davidplowman @naushir

### DIFF
--- a/core/metadata_handler.cpp
+++ b/core/metadata_handler.cpp
@@ -1,0 +1,127 @@
+/* SPDX-License-Identifier: BSD-2-Clause */
+/*
+ * Copyright (C) 2020, Raspberry Pi (Trading) Ltd.
+ *
+ * metadata_hander.cpp - class for handling metadata
+ */
+
+#include "metadata_handler.hpp"
+
+MetadataHandler::MetadataHandler(VideoOptions const *options)
+	: options_(options), still_options_(nullptr), buf_metadata_(std::cout.rdbuf()), of_metadata_(), fmt_(),
+	  first_write_(true), video_mode_(true)
+{
+	if (options != nullptr)
+		fmt_ = options_->metadata_format;
+}
+
+void MetadataHandler::initMetadata(int segmentNum)
+{
+	std::string filename = "";
+	if (video_mode_)
+	{
+		filename = options_->metadata;
+		if (options_->split || options_->segment)
+		{
+			// If in split/segment mode we need to generate filename
+			char subfilename[256];
+			int n;
+			n = snprintf(subfilename, sizeof(subfilename), options_->metadata.c_str(), segmentNum);
+			// Turns the %d into useful numbers
+			if (n < 0)
+				throw std::runtime_error("failed to generate metadata filename");
+			filename = subfilename;
+		}
+	}
+	else
+		filename = still_options_->metadata;
+	of_metadata_.open(filename, std::ios::out);
+	buf_metadata_ = of_metadata_.rdbuf();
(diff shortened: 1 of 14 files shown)


core/metadata_handler.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+#include "metadata_handler.hpp"
+
+MetadataHandler::MetadataHandler(VideoOptions const *options)
+	: options_(options), still_options_(nullptr), buf_metadata_(std::cout.rdbuf()), of_metadata_(), fmt_(),
+	  first_write_(true), video_mode_(true)
+{
+	if (options != nullptr)
+		fmt_ = options_->metadata_format;
+}
```
davidplowman, line 16: I can see how this has all come out of the code that we currently have, but at the same time it does leave some things looking a little awkward. For example, the uncomfortable co-existence of video and still modes.

Thinking out loud a bit, I wonder about removing the `options_` field and replacing it with `video_options_`. Then we have two constructors, one as above that takes a `VideoOptions` (and sets the `video_options_` field) and another that takes a `StillOptions` (and sets the `still_options_` field). We could probably also lose the `video_mode_` flag, as only one of the options is ever set. Would that be a bit neater?

encoder/libav_encoder.cpp
```diff
@@ -353,8 +353,6 @@ void LibAvEncoder::EncodeBuffer(int fd, size_t size, void *mem, StreamInfo const
 	{
 		video_start_ts_ = timestamp_us;
 		segment_start_ts_ = video_start_ts_;
-		pause_state_.pause_timestamp = 0;
-		pause_state_.unpause_timestamp = 0;
```
davidplowman, line 357: Just a bit curious as to why these lines have disappeared...?

encoder/libav_encoder.cpp
```diff
@@ -446,6 +444,37 @@ void LibAvEncoder::initOutput()
...
+		fprintf(fp_timestamps_, "# timecode format v2\n"); // Initialise the file
+	}
+
+	// if recording metadata for files, create file and initialise it
+	if (!options_->metadata.empty())
+	{
+		// The metadata_handler functions take care of segmenting and split when passed the segment number
+		metadata_handler_.initMetadata(segment_num_);
+	}
+
```
davidplowman, line 477: Might this be tidier in a helper function?

encoder/libav_encoder.hpp
```diff
@@ -59,7 +61,7 @@ class LibAvEncoder : public Encoder
 	std::atomic<bool> output_ready_;
 	bool abort_video_;
 	bool abort_audio_;
-	int64_t video_start_ts_;
+	uint64_t video_start_ts_;
```
davidplowman, line 64: Is there a reason for this change?

output/output.hpp
```diff
@@ -47,12 +48,8 @@ class Output
...
 	int64_t time_offset_;
 	int64_t last_timestamp_;
-	std::streambuf *buf_metadata_;
-	std::ofstream of_metadata_;
-	bool metadata_started_ = false;
-	std::queue<libcamera::ControlList> metadata_queue_;
+	int segment_num_;
+	int previous_segment_num_;
+	MetadataHandler metadata_handler_;
+	bool metadata_startedd_;
```
davidplowman, line 54: Double d? And in fact, is it used anywhere?

PR #544, opened 2023-07-25
core: Add flicker correction control

naushir:
Allow user to set the flicker period through the --flicker-period command line argument.


naushir:
This will compile fail the tests as the flicker period control has only just been added to the libcamera tree.

PR #560, opened 2023-08-25
Support libcamera's new SensorConfiguration

davidplowman:
This commit adds support for the new SensorConfiguration within libcamera's CameraConfiguration object. We have to use this now to force the choice of particular sensor modes in case of future extensions where the raw stream format is no longer sufficient.

There are quite a few changes:

1. We now populate the list of sensor modes always, even when no framerate was given. Though we don't fill in the mode's framerate unless that's going to be something we use to select the mode.

2. The selectModeForFramerate function has been generalised and turned into selectMode. This now supports mode selection based on width, height, bitrate as well as (optionally) framerate.

3. Raw streams are always requested now, as this means we always get the same buffer management behaviour. It also simplifies the code slightly.

4. We always fill in a Mode structure (using the user's values if given) that we can pass to the amended selectMode function. This will return the correct sensor mode for us, that we can use to program the SensorConfiguration. So a SensorConfiguration is now always supplied to libcamera.

5. Everywhere where the code was running through the available modes and configuring them, it has been amended to program the SensorConfiguration too - otherwise you may not be able to get non-default bit-depths. To make this easier, the SensorMode class has been made public.

6. The "rawfull" option has been removed. It was only ever a temporary hack from the very early days, and trying to keep it seems to add more complexity that would be warranted.

### DIFF
--- a/core/libcamera_app.cpp
+++ b/core/libcamera_app.cpp
@@ -162,31 +162,41 @@ void LibcameraApp::OpenCamera()
 	post_processor_.SetCallback(
 		[this](CompletedRequestPtr &r) { this->msg_queue_.Post(Msg(MsgType::RequestComplete, std::move(r))); });
 
-	if (options_->framerate)
+	// We're going to make a list of all the available sensor modes, but we only populate
+	// the framerate field if the user has requested a framerate (as this requires us actually
+	// to configure the sensor, which is otherwise best avoided).
+
+	std::unique_ptr<CameraConfiguration> config = camera_->generateConfiguration({ libcamera::StreamRole::Raw });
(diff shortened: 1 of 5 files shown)


davidplowman:
@naushir Not to be merged until SensorConfigurations have been committed into libcamera!

This commit should preserve existing behaviour where it finds the best matching mode when a raw mode is requested, and tries to simplify the configuration process a bit. Obviously it won't compile until those libcamera updates have happened.

davidplowman:
@naushir I think this is actually good now, test failures notwithstanding. Obviously those can't work until they're running on a libcamera that includes sensor configurations.

I've also added a 2nd commit to write DNGs using compressed raw formats as well.

PR #568, opened 2023-09-23
libav: Fix parser with builds where enable_libav=false

sismoke:
this fixes libcamera-vid app when libcamera-apps is built with enable_libav=false option specified

previously the libcamera-vid app exited with the error `Invalid time string provided` no matter what parameters were passed, because the parameter parser was trying to decode parameters that are not present with the `enable_libav=false` build option set

### DIFF
--- a/core/video_options.hpp
+++ b/core/video_options.hpp
@@ -190,10 +190,11 @@ struct VideoOptions : public Options
 		if (Options::Parse(argc, argv) == false)
 			return false;
 
-		av_sync.set(av_sync_);
 		bitrate.set(bitrate_);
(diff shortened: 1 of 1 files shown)


naushir:
Change looks good.  Could you resubmit with a change to the commit message with the following changes please:
- Set the title to "libav: Fix parser with builds where enable_libav=false "
- Add a ``Signed-off-by: ...`` line to the end with your name/email.  Please see other commits for examples of this.

sismoke:
Changed the title and commit message, I hope everything else is all right.

naushir:
Thanks! I'll merge this now.

PR #576, opened 2023-10-05
core: Use correct pipeline config file for different platforms

naushir:
Update and rename check_camera_stack() to get_platform().  This functions now returns the running platform if known. The pipeline config files are now selected based on the detected platform.

As a drive-by add a missing close(fd) when detecting if the HW encoder is present.

### DIFF
--- a/core/libcamera_app.cpp
+++ b/core/libcamera_app.cpp
@@ -25,14 +25,19 @@
 
 unsigned int LibcameraApp::verbosity = 2;
 
-// If we definitely appear to be running the old camera stack, complain and give up.
-// Everything else, Pi or not, we let through.
+enum class Platform
+{
+	UNKNOWN,
+	LEGACY,
+	VC4,
+	PISP,
+};
 
-static void check_camera_stack()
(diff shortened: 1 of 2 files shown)


core/libcamera_app.cpp
```diff
@@ -96,12 +99,19 @@ static void set_pipeline_configuration()
...
 {
-	check_camera_stack();
+	LibcameraApp::Platform platform = get_platform();
+	if (platform == Platform::LEGACY)
+	{
+		// If we definitely appear to be running the old camera stack, complain and give up.
+		// Everything else, Pi or not, we let through.
+		fprintf(stderr, "ERROR: the system appears to be configured for the legacy camera stack\n");
+		exit(-1);
+	}
```
davidplowman, line 109: How about the UNKNOWN case? Might it be better to quit with a helpful message rather than let it run on and for some slightly unspecified failure? (I know it should never happen...)

naushir: Done.

PR #578, opened 2023-10-06
Check several /dev/video nodes in order to find the platform

davidplowman:
There may be other devices pluggin in so /dev/video0 may not reveal the platform correctly. Instead, try the first few /dev/video<n> nodes.


davidplowman:
@naushir

PR #579, opened 2023-10-09
core: Better handling of missing camera dev nodes

naushir:
Preserve the historical behaviour where if a camera dev node is missing, we report it as "no camera available" instead of an unknown platform.


naushir:
@davidplowman maybe better?

PR #581, opened 2023-10-10
core: test: Add a no-raw option

naushir:
Add an command line option "--no-raw" to disable raw stream selection entirely. This is mainly useful for testing the pipeline handler code path that would otherwise not be exercised through libcamera-apps.

Update the test scripts to use this flag in the tests.


naushir:
@davidplowman

PR #582, opened 2023-10-16
core: Add single exposure PiSP HDR mode

naushir:
Update the --hdr option to support the single exposure multi-frame HDR mode available on the PiSP HW. The option can now take the following values:

"off": No HDR
"auto": Automically select the HDR mode, where "sensor" is top priority
"sensor": Sensor HDR (e.g. Camera v3)
"single-exp": Single exposure HDR mode

The implicit value of this argument is "auto".


naushir:
@davidplowman

PR #584, opened 2023-10-17
drm_preview: fix unreleased resources and a typo

Leuca:
In DrmPreview::findCrtc() a series of drmModeGet*() functions is used. Those functions actually allocate memory that needs to be released with the corresponding drmModeFree*() call. In some cases resources are not freed causing a memory leak which can be confirmed with Valgrind.

Free all the allocated resources before losing the pointer.

Also fix a clear typo.


naushir:
Thank you for this fix.  Now merging...

PR #588, opened 2023-10-20
Consider avoiding the use of drm_prime as pixel format

Leuca:
To my knowledge the drm_prime pixel format is something required by the stateless HEVC hardware decoder of the Raspberry Pi.
I can't see any benefit or requirement in using it while using the H264 hardware encoder.

Considering this also has the effect of breaking compatibility with mainline libav, would it be possible to avoid its use?

### DIFF
--- a/encoder/libav_encoder.cpp
+++ b/encoder/libav_encoder.cpp
@@ -22,7 +22,7 @@ namespace {
 
 void encoderOptionsH264M2M(VideoOptions const *options, AVCodecContext *codec)
 {
-	codec->pix_fmt = AV_PIX_FMT_DRM_PRIME;
+	codec->pix_fmt = AV_PIX_FMT_YUV420P;
 	codec->max_b_frames = 0;
 
 	if (options->bitrate)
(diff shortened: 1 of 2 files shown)


naushir:
Indeed, removing this format would break our zero copy v4l2 m2m video encode functionality. Unfortunately we cannot merge this PR.

If you want upstream compatibility, you can use the ``--libav-video-codec libx264`` command line argument to revert to full software encoding.

Leuca:
> DRM prime is an FFmpeg wrapper around dmabufs, so using them should be avoiding a memcpy of the image frames. The naming leaves much to be desired, but the DRM use case got in there first with their naming.
> 
> AV_PIX_FMT_DRM_PRIME is supported by mainline FFmpeg https://github.com/FFmpeg/FFmpeg/blob/master/libavutil/pixfmt.h#L348 It looks like adding the support to v4l2_m2m_enc hasn't made it upstream yet, but there is an effort to get all our patches upstreamed.

Are you sure the h264_v4l2m2m decoder doesn't avoid memcpy of the image frames by itself?

> Indeed, removing this format would break our zero copy v4l2 m2m video encode functionality. Unfortunately we cannot merge this PR.
> 
> If you want upstream compatibility, you can use the `--libav-video-codec libx264` command line argument to revert to full software encoding.

In my tests it seemed like nothing was broken at all...

Is there a way to see that the h264_v4l2m2m is memcpy(ing) the image frames or that you actually loose the zero copy functionality?

naushir:
> Are you sure the h264_v4l2m2m decoder doesn't avoid memcpy of the image frames by itself?

Yes because there is no mechanism to provide the dmabuf/drmprime_buffer file descriptor in through the libav encode API.  You can only provide a framebuffer pointer, meaning non-zero copy operation.

Leuca:
My bad. I didn't realise zero copy v4l2 was involved. I just saw it worked both ways and thought this was due an older version of FFmpeg or the like.

PR #590, opened 2023-10-25
Hdr options changes

naushir:


core/options.cpp
```diff
@@ -104,6 +104,33 @@ static int xioctl(int fd, unsigned long ctl, void *arg)
...
+		v4l2_control ctrl { V4L2_CID_WIDE_DYNAMIC_RANGE, en };
+		xioctl(fd, VIDIOC_G_CTRL, &ctrl);
+		if (ctrl.value != en)
+		{
+			ctrl.value = en;
+			xioctl(fd, VIDIOC_S_CTRL, &ctrl);
+			changed = true;
+		}
+		close(fd);
+	}
```
njhollinghurst, line 129: This will work but it's a bit hard to understand.

For a non-IMX708, the G_CTRL will fail (but that's OK as we don't set the changed flag and will try other devices).
For an IMX708 that's already in use, the S_CTRL may fail (changed may be set pessimistically but it will do no harm).

Is it helpful to test the return values of the ioctls? If not an explanatory comment would be good.

naushir: Good point, I'll make a change to test the ioctl return value.

davidplowman:
It all seems fair enough to me... if quite a lot of effort for something that you feel shouldn't have been needed in the first place!

core/options.cpp
```diff
@@ -197,18 +184,42 @@ bool Options::Parse(int argc, char *argv[])
...
+		libcamera::logSetTarget(libcamera::LoggingTargetNone);
+
+	// HDR control. Set the sensor control before opening or listing any cameras.
+	// Start by disabling HDR unconditionally. Reset the camera manager if we have
+	// actually switched the value of the control.
+	set_subdev_hdr_ctrl(0);
+	app_->initCameraManager();
+
+	// Unconditionally disable libcamera logging for a bit.
+	libcamera::logSetTarget(libcamera::LoggingTargetNone);
```
njhollinghurst, line 200: Does this suppress only infos and debugs, or does it remove errors and warnings?
We might want to keep the latter.

PR #599, opened 2023-11-17
rpicam_still: Add RGB/BGR 48bpp support

naushir:
Add support for 48-bpp RGB/BGR formats that can be generated by the PiSP.  rpicam-still can select this with the "--encoding rgb48" command line option.

### DIFF
--- a/.github/workflows/rpicam-apps-style-checker.yml
+++ b/.github/workflows/rpicam-apps-style-checker.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: [ self-hosted ]
 
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
       with:
         fetch-depth: 0
         clean: true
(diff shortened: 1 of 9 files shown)


naushir:
Cannot really merge this until we have RGB161616 support in libcamera.

schoolpost:
> Cannot really merge this until we have RGB161616 support in libcamera.

coming soon?

schoolpost:
> We will try to get it in our next release. It all depends on how and when we can start the driver upstreaming process. You are welcome to try the change out, but note that you will need the latest kernel and build/install libcamera from https://github.com/raspberrypi/libcamera/tree/next.

Is this format only supported in the Still stream? or also video?

naushir:
The RGB48 format is supported in all configurations.  It's likely you have not updated your kernel, so the driver does not report support for 48-bpp yet, hence your error.

You can either rebuild the kernel yourself, or run ``sudo rpi-update`` to get the latest non-stable version to test out.

PR #609, opened 2023-11-22
Add a Github workflow to generate release tarballs

XECDesign:


naushir:
@XECDesign shall I create a new tag and release with this change included?

PR #615, opened 2023-12-01
Pass app to BufferReadSync by reference

danheron:
rpicam-detect fails to build.

image

The app parameter should be passed by reference.


danheron:
OK, I've updated the commit message.

naushir:
Thanks!  Merging now.

PR #621, opened 2023-12-14
options: Test all variants of IMX708 for HDR mode

naushir:
We were only testing the "imx708" string and ignoring variant suffixes. Fix this by only searching for "imx708" as a substring.


njhollinghurst:
LGTM. Well spotted. (Though on Pi 5 at least, our TDN-based "fake" HDR seems to work as well or better than the sensor's one.)

njhollinghurst (approved):

Jhin4444K:
why my camera auto down
someone have a idea for this situation

Viewfinder frame 99
Closing RPiCam application(frames displayed 80, dropped 19)
Camera stopped!
Tearing down requests, buffers and configuration
Camera closed

PR #634, opened 2024-01-26
libav changes

naushir:


naushir:
@davidplowman can you cast your eyes on this please?  Hopefully should give parity in the command line args between PI 4 and 5 for rpicam-vid.

encoder/libav_encoder.cpp
```diff
@@ -173,17 +173,29 @@ void LibAvEncoder::initVideoCodec(VideoOptions const *options, StreamInfo const
 	// Apply general options.
 	encoderOptionsGeneral(options, codec_ctx_[Video]);
 
-	const char *format;
-	if (options_->output.empty() && options->libav_format.empty())
-		format = "h264";
-	else if (options->libav_format.empty())
-		format = nullptr;
+	const std::string tcp { "tcp://" };
+	const std::string udp { "udp://" };
```
davidplowman, line 177: I'm assuming we don't care about case here!

encoder/libav_encoder.cpp
```diff
@@ -173,17 +173,29 @@ void LibAvEncoder::initVideoCodec(VideoOptions const *options, StreamInfo const
...
+
+	// Setup an appropriate stream/container format.
+	const char *format = nullptr;
+	if (options->libav_format.empty())
+	{
+		if (options->output.empty() ||
+			options->output.find(tcp.c_str(), 0, tcp.length()) != std::string::npos ||
+			options->output.find(udp.c_str(), 0, udp.length()) != std::string::npos)
```
davidplowman: This is slightly awkward to read, I find. Are we using C++20? Apparently that has a `starts_with`...

encoder/libav_encoder.cpp
```diff
@@ -180,9 +180,8 @@ void LibAvEncoder::initVideoCodec(VideoOptions const *options, StreamInfo const
 	const char *format = nullptr;
 	if (options->libav_format.empty())
 	{
-		if (options->output.empty() ||
-			options->output.find(tcp.c_str(), 0, tcp.length()) != std::string::npos ||
-			options->output.find(udp.c_str(), 0, udp.length()) != std::string::npos)
+		if (output_file_.empty() || output_file_.find(tcp.c_str(), 0, tcp.length()) != std::string::npos ||
+			output_file_.find(udp.c_str(), 0, udp.length()) != std::string::npos)
```
davidplowman, line 184: Did the style checker request this change? I think I preferred the original!

encoder/libav_encoder.cpp
```diff
@@ -192,8 +191,17 @@ void LibAvEncoder::initVideoCodec(VideoOptions const *options, StreamInfo const
 	}
 	else
 		format = options->libav_format.c_str();
+
+	// Legacy handling of the --listen parameter.  If missing from the url string, add "?listen=1" to the end.
+	if (options->listen && output_file_.find(tcp.c_str(), 0, tcp.length()) != std::string::npos)
+	{
+		const std::string listen { "?listen=1" };
+		if (output_file_.find(listen, output_file_.length() - listen.length()) == std::string::npos)
```
davidplowman, line 199: Just slightly wondering whether it's necessary to constrain the length of the search here, wouldn't that happen by itself?

davidplowman:
My minor whinging notwithstanding, I'm happy to merge...!

encoder/libav_encoder.cpp
naushir: I think this is case sensitive in libav.

encoder/libav_encoder.cpp
naushir: Unfortunately we are using C++17, but starts_with is exactly what I want :(

encoder/libav_encoder.cpp
naushir: Yes.  I prefer it as the original as well, I'll revert it.

encoder/libav_encoder.cpp
naushir: True, I can get rid of the length constraint.

naushir: Actually, this is the equivalent of ends_with(), so the constraint here is the start position of the search.

PR #637, opened 2024-01-31
rpicam_vid: handling SIGPIPE to avoid sporadic app termination

tederer:
SIGPIPE gets raised when trying to write to an already closed socket. This can happen, when you're using TCP to stream to VLC and the user presses the stop button in VLC. Catching the signal to be able to react on it, otherwise the app terminates.


naushir:
Sorry, one final thing... could you wrap the commit message at 72 lines please?  If you force push to your branch, this will automatically update the PR.

tederer:
It should be there now :-)

Am 2024-01-31 10:23, schrieb naushir:

PR #644, opened 2024-02-19
core: app: Extend the search for compatible video devices

njhollinghurst:
In get_platform(), don't assume that a compatible video device will be found at /dev/video[0-4]. Extending the search range to /dev/video255 is cost-free in cases where it terminates early.


njhollinghurst:
We seem to need this on Pi 5 on Linux 6.7

PR #656, opened 2024-03-05
libav: Warn on using options that don't work with the libav encoder

naushir:
The circular, segment, save_pts and split functions don't work when using the libav encoder.  Loudly warn if the user tries to use one of these options.


naushir:
@davidplowman and @lurch

PR #674, opened 2024-05-04
removing unused cameras_ variable from RPiCamApp

ArosPrince:


naushir:
Thank you for this fix.  The change itself looks fine, but would you be able to change the commit message to follow our convention?  You can see examples in the git log.  Also, we do require a ``Signed-of-by`` tag to be added to the end of the commit message.

ArosPrince:
> Thank you for this fix. The change itself looks fine, but would you be able to change the commit message to follow our convention? You can see examples in the git log. Also, we do require a `Signed-of-by` tag to be added to the end of the commit message.

Hi, this is probably a silly question... But are the conventions described anywhere? I cannot find it anywhere in the repo docs.

I can somewhat reverse engineer what you probably mean by looking at the other repo commit messages, but I wouldn't like to guess...

naushir:
We don't have a template unfortunately.  I would change it to something like:

```
core: Remove unused variable

The cameras_ member variable in class RPiCamApp is unused.
Remove it.

Signed-off-by: My Name <my@email.com>
```

PR #693, opened 2024-07-07
[draft] Selective Recording based on object detection, detection log

matteius:
I want to take a pulse on accepting new features back to the mainline repository.
In my initial exploration of the Hailo AI module, I discovered its possible to record continuously to a file or not at all, however I had a strong desire to trigger recordings based on object detections.

I have a surprising amount of time into these code changes already as I had tried a few approaches that flopped prior.

Current features I've added:
* Added a new RecordingManager class to manage the recording state and trigger recordings based on object detections which is behind an options flag --record-detection <seconds>.  The default is 0 meaning it does not invoke this feature.
* I also generate a detections log file that records the detection and the time they were detected.  This is useful for debugging and understanding the behavior of the object detection.  I should hide this behind an options flag as well.
* Seems that the post-processing run in a separate process and so I use a shared tmp file to communicate recording state between detection process and the actual app.

Current limitations:
* The mp4 containers I generate are not perfect because I start writing to the container as soon as object is detected which is likely before the next key frame.   These play fine in ffplay and vlc, but generally I have been running something like `ffmpeg -i 20240706_074707.mp4 -c copy -map 0 20240706_074707_remuxed.mp4` to clean up the recording files.
* my home directory is hardcoded in at least one spot.
* Only implemented in the rpicam-vid application and the hailo_yolo_inference post processor.

Future work:
* I will likely be doing more in this space, I plant to update the app to have an option of whether to draw the post-processing in the video feed or hide it, as well as put my detection log behind an options flag.
* Would love to generate a more perfect mp4 container with initial key frame+metadata, but I'm not sure how to do that yet.

### DIFF
--- a/post_processing_stages/meson.build
+++ b/post_processing_stages/meson.build
@@ -4,10 +4,6 @@ conf_data = configuration_data()
 conf_data.set('POSTPROC_LIB_DIR', '"' + posproc_libdir + '"')
 configure_file(output : 'postproc_lib.h', configuration : conf_data)
 
-# JSON (and other assets)
-assets_dir = meson.project_source_root() / 'assets'
-postproc_assets = []
-
 # Core postprocessing framework files.
 rpicam_app_src += files([
     'histogram.cpp',
@@ -20,13 +16,6 @@ core_postproc_src = files([
     'hdr_stage.cpp',
     'motion_detect_stage.cpp',
     'negate_stage.cpp',
-])
-
-# Core assets
-postproc_assets += files([
-    assets_dir / 'hdr.json',
-    assets_dir / 'motion_detect.json',
-    assets_dir / 'negate.json',
 ])
 
 core_postproc_lib = shared_module('core-postproc', core_postproc_src,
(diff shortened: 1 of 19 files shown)


post_processing_stages/meson.build
```diff
@@ -103,7 +77,7 @@ endif
 enable_hailo = false
 hailort_dep = dependency('HailoRT', modules : ['HailoRT::libhailort'], version: '>=4.17.0',
                          required : get_option('enable_hailo'))
-hailo_tappas_dep = dependency('hailo-tappas-core', required : get_option('enable_hailo'))
+hailo_tappas_dep = dependency('hailo_tappas', required : get_option('enable_hailo'))
```
matteius, line 80: I forget why I took some of this out of the build the other day but I do know I ran into `hailo-tappas-core` actually being called `hailo_tappas` when I built it from source.

naushir:
@matteius thank you for the effort you have put in to this.  I realise this is still a work in progress, but I just wanted to jump in and discuss a few things before you spend too much time possibly going in a different direction to what we (Raspberry Pi) might want.

Typically for such features, we would prefer not to bake the logic directly into rpicam-apps, but rather do it at the application level.  Of course, this may not always be possible.  The motivation of keeping this feature out of the core ``rpicam-apps`` library is to ensure we minimise the amount of code for application specific logic (like this where it it video specific).

But for what (I think) you are trying to achieve, I think there may be common ground with what we have done in our [rpicam-detect](https://github.com/raspberrypi/rpicam-apps/blob/main/apps/rpicam_detect.cpp)t application.  In there, the application handles the logic to taking a still image capture if a particular object is detected.  Similarly, could we perhaps write an application that initiates a video recording instead (rpicam-vid-detect maybe?).

matteius:
Thanks for the feedback @naushir -- I am pretty busy these days, but I will take a look when I have more time at what the `rpicam-detect` application is doing, as right now the containers I create are playable by ffplay but not well formed enough for my liking, perhaps there are insights in that other application I can gleam for how to create a separate video app that has the feature set I am looking for.

PR #697, opened 2024-07-18
low resolution config changes.

naushir:


naushir:
@davidplowman can you cast your eyes on these changes please?

naushir:
@davidplowman sadly the plan to use ``AdjustConfig`` is not without problems.  ``AdjustConfig`` expects a use case string, rather than a stream identifier string, but I can possibly have a special "lores" use case maybe?

naushir:
Actually, using ``AdjustConfig`` is even more complicated than I expected.  Logically, we should only call ``AdjustConfig("lores")`` if a low res stream is enabled... but the mechanism to enable the low res stream comes through the call to ``AdjustConfig("lores")``.  So that means that the user will still have to write ``--lores-width xx --lores-height xx`` on the command line to get this working.

I think we might be better of leaving ``AdjustConfig`` as is, and using the mechanism in this PR to drive the lo res config.

PR #700, opened 2024-07-22
Update tf_stage.cpp - corrected Inference time display unit

gftrobots:
The inference time (time_taken) is displayed as "ms" which is the standard abbreviation for millisecond, however the variable time_taken is declared in microseconds. The abbreviate for that is "us" if displayed with a standard character set.

auto time_taken = ExecutionTime<std::micro>(&TfStage::runInference, this).count();


naushir:
The change looks good.  However, would you be able to update the commit message to match our standard format?  Have a look at the git log for examples.  You will also need to include a ``Signed-off-by`` tag in the commit message.

PR #704, opened 2024-08-12
core: Better handling of IMX708 sensor HDR

naushir:
Do not set HDR control switch on all IMX708 subdev nodes available. Rather, traverse through sysfs and find the correct subdev node to toggle the HDR control on and off. We aim to match the selected camera device id when deciding which subdev node to use.

### DIFF
--- a/core/options.cpp
+++ b/core/options.cpp
@@ -6,6 +6,7 @@
  */
 #include <algorithm>
 #include <fcntl.h>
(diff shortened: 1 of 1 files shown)


njhollinghurst:
LGTM. I see we now check specifically for "imx708" which seems fine as other sensors might set HDR modes differently and we have no other examples.

PR #782, opened 2025-02-27
core: Set the lores colourspace to match the main stream colourspace.

naushir:
This matches the behaviour of Picamera2.


davidplowman (approved):
Yep, not a lot here one could argue with!!

PR #792, opened 2025-04-01
libav: Use avcodec_get_supported_config instead of deprecated struct …

r41k0u:
Since https://ffmpeg.org/pipermail/ffmpeg-devel/2024-April/325227.html, accessing the `sample_fmts` struct field has been deprecated. This change replaces that with an `avcodec_get_supported_config` call.


r41k0u:
Hi, yes I saw the failures. I tried this using ffmpeg 7.1.1 and 7.0
This change will allow the bump to ffmpeg 7.1.1 (the existing code works with 7.0 without deprecation warnings)

naushir:
Unfortunately we won't be able to merge this until we bump up our ffmpeg library version.  I'll close this one now and we will revisit when we have the correct dependencies.

naushir:
https://github.com/raspberrypi/rpicam-apps/pull/800 adds and ``#ifdef`` that will conditionally compile this with our existing ffmpeg version on Raspberry Pi OS.

PR #813, opened 2025-06-04
Couple of changes

naushir:


davidplowman:
> @davidplowman do you think the world is ready for [7c80896](https://github.com/raspberrypi/rpicam-apps/commit/7c80896becc10e1d0cf30e853fdc42571eb03def)?

Seems fair enough to me! Though a message on the forum making it very obvious that this will happen in the next release would probably not go amiss...

PR #827, opened 2025-06-21
core: Update checks for disable_rpi_features option

bkuhls:
Based on the original commit that added the sync feature there were a few more options that needed the disable_rpi_features code conditionals to ensure compilation against upstream libcamera so add them in to ensure this works.

Fixes:
```
../core/rpicam_app.cpp: In member function ‘void RPiCamApp::StartCamera()’: ../core/rpicam_app.cpp:642:78: error: ‘controls::rpi’ has not been declared
  642 |         if (!controls_.get(controls::ScalerCrop) && !controls_.get(controls::rpi::ScalerCrops))
      |                                                                              ^~~
../core/rpicam_app.cpp:673:49: error: ‘controls::rpi’ has not been declared
  673 |                         controls_.set(controls::rpi::ScalerCrops, libcamera::Span<const Rectangle>(crops.data(), crops.size()));
      |                                                 ^~~
```

### DIFF
--- a/core/rpicam_app.cpp
+++ b/core/rpicam_app.cpp
@@ -637,6 +637,7 @@ void RPiCamApp::StartCamera()
 	// This makes all the Request objects that we shall need.
 	makeRequests();
 
+#ifndef DISABLE_RPI_FEATURES
 	// Build a list of initial controls that we must set in the camera before starting it.
 	// We don't overwrite anything the application may have set before calling us.
 	if (!controls_.get(controls::ScalerCrop) && !controls_.get(controls::rpi::ScalerCrops))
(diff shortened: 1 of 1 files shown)


naushir:
controls::rpi::ScalerCrops is available [upstream](https://git.libcamera.org/libcamera/libcamera.git/tree/src/libcamera/control_ids_rpi.yaml#n35), so I'm not sure that I would cover this with the ``disable_rpi_features`` option.  Is there a reason you need to do this?

bkuhls:
This patch fixes a build error for non-rbpi platforms.
Please note that I stumbled across this build error while updating the buildroot package from version 1.5.0 to [1.7.0](https://gitlab.com/buildroot.org/buildroot/-/commit/c6252c2e50b7fe555f62cbf38fdde8ba540f00ea) to prepare a ffmpeg version bump. Beyond that I did not even knew this package, just trying to fix a build error ;) Buildroot does, for whatever reason, not restrict the build of this package for rbpi platforms.

naushir:
It seems like libcamera is built without Raspberry Pi support.  I suggest you enable Raspberry Pi support in libcamera, and then building ``rpicam-apps`` will not cause an error.

To be honest, I doubt ``rpicam-apps`` would work on any other platform.

PR #829, opened 2025-06-23
core: Hide startup frames

naushir:
Hide frames that are not marked with FrameMetadata::FrameSuccess, as they could be either invalid or startup frames.


davidplowman:
What a big change! LGTM...

PR #884, opened 2025-11-27
Adding monochrome DNG support

will127534:
This is for adding monochrome support for DNG raw capture.

- Added a new flag in BayerFormat for monochrome
- Use that added flag to branch out differences between CFA RAW DNG and monochrome (Linear DNG)

Added support for:

- R10_CSI2P
- R12_CSI2P
- R16
- MONO_PISP_COMP1

Tested with IMX585 monochrome on RPI5 with both :U (R16) and :P (MONO_PISP_COMP1)

### DIFF
--- a/image/dng.cpp
+++ b/image/dng.cpp
@@ -20,6 +20,10 @@
 #define MAKE_STRING "Raspberry Pi"
 #endif
 
+#ifndef PHOTOMETRIC_LINEARRAW
+#define PHOTOMETRIC_LINEARRAW 34892
+#endif
+
 using namespace libcamera;
 
 static char TIFF_RGGB[4] = { 0, 1, 1, 2 };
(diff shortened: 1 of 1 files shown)


naushir:
@will127534 I've reworked this commit slightly, I think there should be more separation of bayer and mono formats, since they are not technically the same.  Can you have a quick test of commit https://github.com/raspberrypi/rpicam-apps/commit/34572c469c99d9a41f018ef9cd242ff35fb5958f please?

will127534:
Hi @naushir thanks a lot! 
I can confirm that the code is working with the same setup and both  :U (R16) and :P (MONO_PISP_COMP1) formats.

I was trying to keep the changes as few as possible so it is easier to review and merge but this is the correct way. 
On the same notes, can we remove this CCM missing warning for mono also?
https://github.com/raspberrypi/rpicam-apps/blob/34572c469c99d9a41f018ef9cd242ff35fb5958f/image/dng.cpp#L418

will127534:
Thank you so much!

Issue #885, opened 2025-12-05
[BUG] rpicam-jpeg --metadata option does not create metadata file


Kletternaut:
# rpicam-jpeg --metadata option does not create metadata file

## Problem

`rpicam-jpeg --metadata` is documented but does not work. The option is parsed but no metadata file is created.

## Reproduce

```bash
rpicam-jpeg -o test.jpg \
  --metadata test_metadata.json \
  --metadata-format json \
  -t 2000 \
  --qt-preview
```

**Result:** Image created, metadata file NOT created.

`rpicam-still` with same parameters works correctly.

## Root Cause

`save_metadata()` function is never called in `apps/rpicam_jpeg.cpp`.

## Fix

```diff
--- a/apps/rpicam_jpeg.cpp
+++ b/apps/rpicam_jpeg.cpp
@@ -8,6 +8,8 @@
 #include "core/rpicam_app.hpp"
 #include "core/still_options.hpp"
 
+#include "output/output.hpp"
+
 #include "image/image.hpp"
 
 using namespace std::placeholders;
@@ -29,6 +31,22 @@ public:
 	}
 };
 
+static void save_metadata(StillOptions const *options, libcamera::ControlList &metadata)
+{
+	std::streambuf *buf = std::cout.rdbuf();
+	std::ofstream of;
+	const std::string &filename = options->Get().metadata;
+
+	if (filename.compare("-"))
+	{
+		of.open(filename, std::ios::out);
+		buf = of.rdbuf();
+	}
+
+	write_metadata(buf, options->Get().metadata_format, metadata, true);
+}
+
 // The main even loop for the application.
 
 static void event_loop(RPiCamJpegApp &app)
@@ -84,6 +102,8 @@ static void event_loop(RPiCamJpegApp &app)
 			BufferReadSync r(&app, payload->buffers[stream]);
 			const std::vector<libcamera::Span<uint8_t>> mem = r.Get();
 			jpeg_save(mem, info, payload->metadata, options->Get().output, app.CameraModel(), options);
+			if (!options->Get().metadata.empty())
+				save_metadata(options, payload->metadata);
 			return;
 		}
 	}
```

## Test Results

Tested with JSON, TXT, and stdout formats - all working correctly.

naushir:
This fix looks reasonable.  Would you be able to post a PR with this change?  If not, I can do it.

PR #887, opened 2025-12-11
Fix rpicam-jpeg --metadata option

Kletternaut:
The --metadata option was documented but not functional in rpicam-jpeg. While rpicam-still correctly saves metadata files, rpicam-jpeg never called the save_metadata() function.

This commit adds:
- Include for output/output.hpp
- save_metadata() function implementation
- Call to save_metadata() after jpeg_save()

Tested with JSON, TXT, and stdout formats - all working correctly.

Fixes: #885

### DIFF
--- a/apps/rpicam_jpeg.cpp
+++ b/apps/rpicam_jpeg.cpp
@@ -10,6 +10,8 @@
 #include "core/rpicam_app.hpp"
 #include "core/still_options.hpp"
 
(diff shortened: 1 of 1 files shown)


Kletternaut:
Done! I've added the sign-off to the commit message.

Kletternaut:
Thanks @naushir — great, thanks for fixing that manually. Much appreciated!

PR #916, opened 2026-05-22
preview: Attempt fallback if Qt preview unavailable

njhollinghurst:
Previously it could crash in the (rare) case that Qt preview was requested but could not be found; there was no path to fall back to the null preview. Now it will continue to search for EGL, DRM or null preview. Tidied code a little.


njhollinghurst:
using `std::vector<std::string>` rather than Dequeue or whatever, because it's already linked and thus less likely to lead to binary bloat.

